PR #14, opened 2026-04-20
Migrate LSP integration to Serena + ty

alimanfoo:
## Summary

- Wire up a language-server MCP for cross-file operations (find references, rename, check usage) on top of uncoded's existing grep-free navigation via namespace map and stubs.
- Bridge: [oraios/serena](https://github.com/oraios/serena). Backend: [astral-sh/ty](https://github.com/astral-sh/ty). Launched via `uvx` — no global install needed.
- Inject Serena-aware cross-file guidance into uncoded's generated navigation section (`_SECTION_BODY`), so agents working in any uncoded-indexed repo know to prefer LSP tools over grep, with concrete tool names and name-path syntax.
- Document the setup in README under *Using uncoded with a language server*.
- Commit `.claude/settings.json` to auto-enable Serena and allowlist its navigation + symbol-edit tools for Claude Code.
- Add `resources/lsp-research.md` capturing the LSP-for-agents landscape as of April 2026.

## History of this PR

This PR began as an `mcp-language-server` + pyright integration, then pivoted to Serena + ty after investigation turned up concerns about `mcp-language-server`'s maintenance health — see the first design comment below. The pivot is now complete.

## Test plan

- [x] Serena read tools (`find_symbol`, `find_referencing_symbols`, `get_symbols_overview`) respond correctly against this repo.
- [x] Reference parity with pyright: `is_public` and `walk_source` return the same use-site references through ty as they did through pyright. Serena's API correctly omits module-level import lines, which lowers the total count from 14→13 and 10→8 respectively but the underlying reference data matches.
- [x] Rename works cross-file: renamed `walk_source` to `walk_source_ACC_TEST`, updated all 11 expected locations across `extract.py` / `cli.py` / `test_extract.py`; 114 tests pass after rename; reverted via git. Rename initially also followed imports into `.uncoded/stubs/` — fixed by adding `.uncoded` to Serena's `ignored_paths`.
- [x] Regenerate namespace map / stubs / CLAUDE.md / AGENTS.md via \`uv run uncoded\`.
- [x] \`uv run pre-commit run --all-files\` passes (ruff, ty, uncoded).
- [ ] Second reviewer to sanity-check the Serena + ty recommendation and the committed allowlist.

## Notes for reviewer

- **Diagnostics are N/A on Serena.** Serena does not expose a diagnostics tool at the MCP level — this is a deliberate upstream design, not a ty limitation. The generated nav guidance no longer promises diagnostics.
- **Pyright config is gone.** \`[tool.pyright]\` was removed from \`pyproject.toml\`: ty handles src-layout repos natively, so the \`venvPath\` / \`extraPaths\` workaround is unnecessary. To fall back to pyright, switch \`languages: ["python_ty"]\` to \`["python"]\` in \`.serena/project.yml\` and restore the \`[tool.pyright]\` block from git history.
- **First-run mutation of \`.serena/project.yml\`.** On its first launch, Serena rewrites the project config with its full template, preserving the values we set. This is expected; downstream users should not be alarmed by the post-launch diff.
- **Committable \`.claude/settings.json\`.** Enables the Serena server and allowlists read-only navigation + rename + structural symbol edits + memory tools. \`open_dashboard\` remains prompt-on-use (it opens a browser). Memory tools are allowlisted because memories live in \`.serena/memories/\` — local Serena state, not repo code.


alimanfoo:
## Context for a future agent: migrate to Serena + ty when picking this up

This PR is left as **draft** because an investigation during the session turned up concerns about the `isaacphi/mcp-language-server` project's maintenance health. This comment is a self-contained brief so the next agent doesn't have to redo the diligence.

---

### TL;DR

**Prototype Serena (`oraios/serena`) as the MCP-LSP bridge, with ty (`astral-sh/ty`) as the Python language-server backend. Fall back to pyright if ty fails acceptance tests.**

---

### Why this PR is draft rather than merge-ready

1. **The current bridge (`isaacphi/mcp-language-server`) is drifting.**
   - 1,512 stars, but last non-bot commit **2025-05-16** (~11 months). Still at **v0.1.1** — never hit stable.
   - Maintainer's own statement, issue #92 (Feb 2026): *"Unfortunately, I don't have too much time to maintain this repo at the moment. I've found that it's quite difficult to maintain due to how differently various language servers behave."*
   - Open issues on core capabilities: **#104** rename limited to single file; **#121** symbols never found (matches dotted-name issue we hit); **#86** inconsistent references by symbol kind; **#83** too many open files.

2. **The tool's own schema is misleading.** It advertises `'mypackage.MyFunction', 'MyType.MyMethod'` as valid symbol-name formats, but only bare single identifiers work. The committed `_SECTION_BODY` in this PR says "use bare names" — accurate, but a quality smell nonetheless.

Neither is a blocker for what we committed — everything in this PR works — but both argue against building more on this foundation.

---

### Recommended target stack

**Serena + ty.** Why each:

**Serena over alternatives**, from `resources/lsp-research.md`:
> *"Serena is the most-adopted bridge across Claude Code, Cline, Cursor, Roo Code, and Codex CLI."*

Comparison (both captured 2026-04-21):

| | `isaacphi/mcp-language-server` | `oraios/serena` |
|---|---|---|
| Stars | 1,512 | **23,187** (~15×) |
| Last push | 2026-03-01 (bot) | **2026-04-20** (1 day before this PR) |
| Model | Raw LSP passthrough | Symbol-path addressing via `multilspy` |
| License | BSD-3 | MIT |

Architectural advantages:
- **Symbol-path addressing** (`ClassName/methodName`) replaces the bare-vs-dotted ambiguity by design.
- **Lifecycle reasoning** (lazy init, incremental indexing, background queue) — directly relevant to issue #83 on the current bridge.
- **Token-efficient symbol-level responses** rather than verbose raw LSP dumps.

**ty over pyright/jedi as the Python backend.** ty is both a type checker and a language server despite common assumption otherwise; the README is explicit:
> *"An extremely fast Python type checker **and language server**, written in Rust."*

Per [docs.astral.sh/ty/features/language-server](https://docs.astral.sh/ty/features/language-server/), ty's LSP supports `textDocument/references`, `textDocument/rename`, `textDocument/definition`, `textDocument/hover`, and `textDocument/diagnostic` — all four capabilities we'd actually use.

Reasoning for ty-first despite beta status:
- **Astral's track record** on uv and ruff raises the expected quality of a beta from the same team. Not a blank cheque, but a meaningful prior.
- **10–100× performance over pyright/mypy** matters more under agent use than human editing — every LSP call's latency compounds across a task.
- **If ty works, we skip two fallback iterations.** Swapping Serena's Python backend is a one-line config change, so the "fail fast" cost of trying ty first is ~15 minutes, not a session.

Serena supports ty natively via a `ty_server.py` in `src/solidlsp/language_servers/` and a `ls_config.py` enum. No integration work needed at the Serena layer.

---

### Acceptance criteria for "ty works"

**Do not rely on Astral's docs table alone.** Pyright's docs also claim rename is supported, and `resources/lsp-research.md` documents that claim failing in practice (Sublime reporting *"Nothing to rename"*, nvim issue #1985 on rename not propagating, etc.). Concrete empirical tests on this repo:

- **References parity.** `is_public` should return **14 references** (6 in `src/uncoded/extract.py`, 8 in `tests/test_extract.py`). `walk_source` should return **10 references** (2 in `src/uncoded/cli.py`, 8 in `tests/test_extract.py`). These baselines were verified on the current mcp-language-server + pyright stack in this session — matching them on Serena + ty is the minimum.
- **Rename edge cases.** Create a symbol with cross-file references, an import alias, and a same-name shadow in a different module. Confirm rename updates all real callers, doesn't touch the shadow, and preserves the alias. This is the canonical pyright failure mode.
- **Diagnostics against a broken file.** Introduce a type error in a scratch file, confirm ty surfaces it, revert. (This was never tested even on pyright — worth doing either way.)
- **Stability smoke test.** Run a series of edits + reloads over ~10 minutes. Watch for crashes, stale diagnostics, sync drift.

If ty fails any of these, fall back in this order:

1. **Serena + pyright** (Serena's default) — the `[tool.pyright]` config committed in this PR carries over unchanged; pyright reads its own config regardless of launcher.
2. **Serena + jedi** (`python_jedi` language key) — only if rename is a target capability and pyright fails it. *Caveat: jedi's "better rename" reputation is general knowledge, not independently verified. Empirical test before recommending.*

---

### Already verified — don't redo

- ✅ On the current `isaacphi/mcp-language-server` + pyright stack: `hover`, `definition` (bare name), `references` (bare name), and `diagnostics` all respond correctly. Dotted symbol names do **not** work in any form tested (`pkg.mod.fn`, `mod.fn`, `Class.method`, `Class.attr`) — this is a property of `mcp-language-server`'s symbol-lookup, not a pyright config issue.
- ✅ Pyright needs **three** config keys for cross-file references from `tests/` to work on a src-layout repo:
  - `venvPath` + `venv` — resolve deps and the package itself via the `.venv`.
  - `extraPaths = ["src"]` — treat `src/` as first-party. Without this, pyright classifies the editable-install package as third-party and does not link cross-file references into it.
  - Detailed rationale in the `pyproject.toml` `[tool.pyright]` section's inline comment.
- ✅ `uv sync` must be re-run if the repo checkout path changes — the `_editable_impl_uncoded.pth` file in `.venv/lib/...` contains an absolute path. A stale `.pth` from a moved checkout silently breaks references with confusing symptoms (some symbols found, others not).

---

### Pyright's rename ceiling is pyright's, not the bridge's

If staying on pyright (for any reason), **drop `rename_symbol` from `_SECTION_BODY`'s recommended tool list** as an honest reflection of pyright's limits, independent of bridge choice. Evidence from `resources/lsp-research.md`:
> *"pyright is the exception that proves the rule. pyright's rename and find-references support is unreliable: Sublime LSP-pyright reports 'Nothing to rename'; nvim-lspconfig issue #1985 documents vim.lsp.buf.rename() not propagating across files; kakoune-lsp #452 shows pyright returning null for rename."*

Our `references` test worked in this session, but the research doc's warning should still inform recommendations — the successful run was narrow.

---

### Other caveats worth knowing

- **ty is beta.** `0.0.x` versioning; the README warns *"breaking changes, including changes to diagnostics, may occur between any two versions."* Pin a specific ty version in the project config and budget revisiting it.
- **Serena has a shell-exec tool** that prompted a 2025 community audit (Serena discussion #380). Configure Serena to disable `execute_shell_command` if shell access is undesirable — Serena's own config system supports per-tool disabling.
- **Docs-say-supported ≠ tested-and-works.** Both the pyright-unreliable evidence and the mcp-language-server schema overclaim are instances of this. Empirical verification first; documentation trust after.
- **Tool-surface cognitive cost.** Serena exposes a richer tool vocabulary (`find_symbol`, `find_referencing_symbols`, `replace_symbol_body`, `insert_after_symbol`, etc.) than the five we use today. The `_SECTION_BODY` guidance will need to be rewritten for these tool names and for Serena's name-path syntax, not just swapped one-for-one.

---

### Starting points

- `resources/lsp-research.md` — full April-2026 landscape, trustworthy on positioning claims.
- `.mcp.json` — current MCP server config. Replace with Serena's launch command (uvx-based per Serena's README).
- `src/uncoded/instruction_files.py` — `_SECTION_BODY` contains the LSP guidance injected into CLAUDE.md/AGENTS.md. Rewrite the "For cross-file operations" paragraph for Serena's tool names and symbol-path addressing. Drop `rename_symbol` from the list unless ty confirms reliable Python rename.
- `README.md` "Using uncoded with a language server" section — needs parallel update.
- `pyproject.toml` `[tool.pyright]` — keep or remove depending on final backend choice; retain if any pyright fallback path is documented.
- This PR's commit `d626f6a` — shows the pyright config reasoning step-by-step (venvPath-alone didn't work; extraPaths was needed for first-party treatment). Useful as a worked example of how pyright config interacts with src-layout repos.

---

### Revision history

This brief reflects three refinements during the original session:
1. Initial recommendation framed as "migrate to Serena." Reviewer pushback (*"does Serena actually use pyright?"*) uncovered that Serena supports multiple Python backends (pyright default, jedi, ty) — the pyright-is-the-ceiling argument needed narrowing.
2. Second pushback (*"ty is a type checker, not a language server"*) prompted verification — ty is both, and Astral's docs claim all four LSP methods we use are supported.
3. Third reset (*"if ty works, that's where to start"*) flipped the prototyping order from conservative-default to ty-first, weighted by Astral's track record on prior projects.

Earlier iterating comments on this PR were removed in favour of this consolidated brief. The reasoning path is preserved here for transparency; the underlying evidence is in `resources/lsp-research.md` and the tool docs linked above.

PR #16, opened 2026-04-22
Add 'check' subcommand as a zero-mutation freshness gate

alimanfoo:
## Summary

- Adds `uncoded check`: runs the full pipeline, writes nothing, exits 1 if any artifact would change. Complements the existing pre-commit modify-then-fail hook for CI pipelines that must not mutate the tree.
- Restructures the CLI into explicit subcommands (`sync`, `check`, `setup-serena`) — bare `uncoded` no longer silently builds; it errors with a usage hint. This eliminates the `--check`+subcommand ambiguity raised in review.
- Refactors writes through a single `sync_file` helper so apply mode only rewrites files whose content actually changed (quieter output, cleaner mtimes).
- Closes the CLI-has-no-direct-tests gap tracked in `MAINTENANCE.md`: new `tests/test_cli.py` covers argparse routing, dispatch, error paths, and check-mode drift detection.

## Why

Staleness was a silent failure mode. Editing a source file without re-running `uncoded` leaves the next agent (or CI) reading stubs with stale line ranges and signatures — and there was no way to verify freshness without mutating the working tree. `check` gives CI a clean gate.

## Breaking change

Per `CLAUDE.md`, alpha repo, no backwards-compat constraint. Migration:
- `uncoded` → `uncoded sync`
- pre-commit `entry: uncoded` → `entry: uncoded sync`

README, CLAUDE.md, AGENTS.md, this repo's own `.pre-commit-config.yaml`, and the Serena project.yml comment are all updated.

## Test plan

- [x] `uv run pytest` — 160 pass (+39 new: sync helpers, per-layer check mode, CLI dispatch end-to-end including unknown/missing subcommands).
- [x] `uv run pre-commit run --all-files` — all hooks pass.
- [x] Manual drift test: prepended a function to `src/uncoded/sync.py`, confirmed `uncoded check` reports the stub + namespace-map as out of date and exits 1; restored, confirmed clean exits 0.
- [x] Confirmed apply mode is now quieter on clean rebuilds (no per-file prints when content unchanged).


alimanfoo:
Thanks for the review. On independent reflection I agree with both points: `--check` as a flag on a silent default invited the ambiguity (what does `setup-serena --check` mean?), and once `check` is a subcommand, the build needs to be one too for parity.

Pushed `3662265`:
- `uncoded sync` — build/refresh (was bare `uncoded`)
- `uncoded check` — verify, no writes (was `uncoded --check`)
- `uncoded setup-serena` — unchanged
- Bare `uncoded` now errors with a usage hint via `subparsers.required=True`

Naming chose `sync` over `build` because the operation reconciles (deletes orphan stubs); `build` would understate it. `check` parallels `ruff check` / `pre-commit check`.

Migration is one-line in pre-commit configs (`entry: uncoded` → `entry: uncoded sync`) and that's it. Alpha repo, no compat constraint. README, CLAUDE.md, AGENTS.md, and the in-repo pre-commit hook all updated.

PR #45, opened 2026-04-30
[claude] Rename setup-serena to setup

### LINKED ISSUE #44: Consider renaming setup-serena to just setup
It's aesthetically simpler and easier to type and remember, `uncoded setup`. The details of what setup does are less relevant to expose.

alimanfoo:
## Summary

Closes #44.

- Renames the user-facing CLI subcommand `setup-serena` → `setup`. Implementation details (Serena, ty, Claude Code) are no longer exposed at the surface; `uncoded setup` is shorter and easier to remember.
- Aligns the internal vocabulary with the new surface: function `setup_serena` → `setup`, test class `TestSetupSerena` → `TestSetup`. Module file `serena_setup.py` and test file `tests/test_serena_setup.py` are deliberately retained — every constant in the module is `SERENA_*`-prefixed and the module's centre of mass remains the Serena integration.
- Moves `read_project_name` from `serena_setup.py` into `config.py` so all `read_*` functions that read from `pyproject.toml` live together (per coherence review S3 in `.uncoded/reviews/2026-04-25-001215.md`).
- Updates docstrings, comments, and assert messages that still referenced the old `setup-serena` subcommand name. The historical review report under `.uncoded/reviews/` is intentionally left as-is.

## Test plan

- [x] `uv run pytest` — 170 passed
- [x] `uv run pre-commit run --all-files` — all hooks green (ruff, ruff format, ty, uncoded)
- [x] `uncoded --help` shows `{sync,check,setup}`
- [x] `uncoded setup` runs successfully against a fresh project
- [x] `uncoded setup-serena` errors with `invalid choice`
- [x] Repo-wide grep for `setup-serena` returns zero hits in tracked source/tests/docs (only the historical timestamped review report still mentions the old `setup_serena` *symbol* names, which is intentional)


alimanfoo:
# Review of PR #45 — "Rename setup-serena to setup"

**Recommendation: approve / merge as-is.** This is a tight, well-scoped rename. Three commits, each leaving the codebase coherent on its own; the diff lines up with the description; the test suite is green (170 passed) and `pre-commit run --all-files` is clean. CLI behaviour matches what the PR description claims (`uncoded --help` shows `{sync,check,setup}`; `uncoded setup-serena` exits with `invalid choice`). I checked the tracked tree for stale `setup-serena` / `setup_serena` mentions and the only hit is the historical timestamped review under `.uncoded/reviews/`, which the description correctly flags as intentionally preserved. Nothing blocking; a few small observations below for the record.

## Strengths worth naming

- **Commit shape is right.** `accd2d6` does the surface rename, `e451087` mops up stale strings, `3883cfc` does the symbol rename + `read_project_name` move. Each commit is independently coherent and easy to bisect.
- **The `read_project_name` move is the right call.** Lifting it from `serena_setup.py` into `config.py` puts all `read_*` pyproject readers next to each other (`find_pyproject_toml`, `read_project_name`, `read_source_roots`, `read_instruction_files`) and removes `tomllib` from `serena_setup.py` — that module no longer parses TOML at all, which is a nice tightening of responsibility. The associated test class moved to `tests/test_config.py` to match.
- **Decision to keep `serena_setup.py` / `tests/test_serena_setup.py` as filenames is defensible and called out.** Every constant in the module is `SERENA_*`-prefixed and the file is genuinely about Serena integration; renaming the module would have churned imports without making the substance more accurate. Sensible restraint.
- **Stubs and `namespace.yaml` were regenerated to match** — `read_project_name` now appears under `config.py`, `setup` (not `setup_serena`) under `serena_setup.py`, `TestSetup` and `TestReadProjectName` in the right test files. The dogfooding pre-commit hook would have caught any drift here, and it's clean.
- **Doc/comment updates went deep enough.** The pin comment in `serena_setup.py:32-36`, the `_sync_mcp_json` docstring, the `TestRepoDogfooding` docstring, and the assert message at `tests/test_serena_setup.py:170` all swap "setup-serena" → "`uncoded setup`". These are the kinds of strings rename PRs usually miss; this one didn't.
- **No backwards-compat alias, and that's correct here.** `SYSTEM.md:31-33` ("Backwards compatibility is not a concern at this stage") makes a deprecation shim actively wrong; clean break is the right call.

## Non-blocking observations

- **`from uncoded.serena_setup import setup` reads slightly oddly** in `src/uncoded/cli.py:11` and the test imports — module name is more specific than the function name. The PR description anticipates this trade-off and lands on the right side of it (the module's centre of mass is Serena), but worth being aware that any future second `setup` (e.g. for a different LSP backend) would need either a rename or qualified imports. Not something to address now.
- **`tests/test_serena_setup.py::TestSetup::test_falls_back_to_cwd_name_when_no_pyproject` (line 143-147) is now partly redundant** with `tests/test_config.py::TestReadProjectName::test_falls_back_to_cwd_name_when_no_pyproject`. It still has independent value as an end-to-end check that `setup()` actually wires `read_project_name()` into `project.yml`, so I wouldn't remove it — just noting the small overlap that the move surfaced. If you wanted to make the integration intent explicit, the test could be renamed to something like `test_uses_cwd_name_in_project_yml_when_no_pyproject`. Minor.
- **PR title vs commit subject.** The PR is titled "Rename setup-serena to setup" but is actually two coupled changes: the rename plus the `read_project_name` relocation. The body calls both out clearly, so this is fine, but the title slightly undersells the second change.

## Nits

- `tests/test_serena_setup.py:7-11` — the `from uncoded.serena_setup import (...)` block now imports just three names (`SERENA_ALLOWED_TOOLS`, `SERENA_VERSION`, `setup`); could be a single line, but the parenthesised form leaves room to grow and is consistent with the rest of the file. Take it or leave it.
- `src/uncoded/serena_setup.py:36` — the comment `A dogfooding test in tests/test_serena_setup.py guards against drift.` is still accurate post-rename. Good.

## Out of scope (deliberately not flagged)

I noticed one or two pre-existing items elsewhere in the tree, but per your remit I'm holding the line on this PR's diff only. Happy to write them up separately if useful.

Issue #52, opened 2026-04-30
[claude] setup() root parameter has no CLI flag


alimanfoo:
The `setup` function exposes a `root` parameter:

- `src/uncoded/serena_setup.py` — `def setup(root: Path | None = None) -> int:`

But the CLI entry point (`src/uncoded/cli.py::main`) calls it with no arguments:

```python
if args.command == "setup":
    return setup()
```

So `root` is reachable from tests but never from the command line — only `Path.cwd()` is exposed as the destination for the generated config files.

This is one of:

- **Test scaffolding** — the parameter exists only so tests can point at a tmp_path. If so, a comment near the signature should make the intent clear, or the test seam should be factored differently (e.g. tests `os.chdir` rather than passing `root`).
- **Unexposed feature** — the parameter implies "you can run `setup` against an arbitrary directory" but the CLI doesn't surface it. If so, an `--root <path>` flag would let users actually use it.

Surfaced by the developer agent in the post-merge sweep for PR #45. Pre-existing.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #51, opened 2026-04-30
[claude] cli.main dispatch falls through to _sync for any non-setup command


alimanfoo:
The dispatch in `src/uncoded/cli.py::main` ends with (paraphrased):

```python
args = parser.parse_args()

if args.command == "setup":
    return setup()
return _sync(check=args.command == "check")
```

This is correct *today* because argparse pre-validates `args.command` against the registered subparsers (`{sync, check, setup}`), so the only commands that can reach the fall-through are `sync` and `check`. But the structure is a maintenance hazard: a future fourth subcommand silently routes to `_sync` unless someone remembers to add an `elif`.

Refactor options worth considering:

- Use `subparser.set_defaults(func=...)` on each subparser and dispatch via `args.func(args)` — eliminates the dispatch ladder entirely.
- An exhaustive `match` statement with no default clause.
- An explicit `elif args.command in ("sync", "check"): return _sync(...)` to remove the silent fall-through.

Surfaced by the developer agent in the post-merge sweep for PR #45. Pre-existing; the rename in PR #45 didn't introduce or worsen the fragility, but the developer was sitting with this code through tasks #1 and #3 and noticed it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #50, opened 2026-04-30
[claude] _sync_serena_project is named like a sync but only writes when absent


alimanfoo:
The three template-writing helpers in `src/uncoded/serena_setup.py` are named in parallel — `_sync_mcp_json`, `_sync_serena_project`, `_sync_claude_settings` — but only two of them actually sync.

- `_sync_mcp_json` merges the Serena entry into existing `.mcp.json` content, refreshing it on every run.
- `_sync_claude_settings` merges the Serena allowlist into existing `.claude/settings.json` content.
- `_sync_serena_project` returns `"unchanged"` immediately if the file exists. It only writes when absent.

The third function's "do not clobber hand-edited Serena config" behaviour is intentional and is documented in the module docstring (`src/uncoded/serena_setup.py:23-24`). But the `_sync_*` naming implies merge / refresh semantics that this function deliberately rejects, forcing a reader of the trio to do extra reconciliation between name and body.

Possible directions:

- Rename to `_write_serena_project_if_absent` (or similar) so the differing behaviour is visible at the call site.
- Or add a one-line comment at the function definition explaining the deliberate departure from the sync semantic.

Surfaced by the developer agent in the post-merge sweep for PR #45. Pre-existing; not introduced or worsened by the rename.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #49, opened 2026-04-30
[claude] config.py module docstring drifted after read_project_name move


alimanfoo:
After commit `3883cfc` (PR #45) moved `read_project_name` from `serena_setup.py` into `config.py`, the `config.py` module docstring no longer accurately describes the file's contents.

- `src/uncoded/config.py:1` — `"""Read uncoded configuration from pyproject.toml."""`
- `read_project_name` reads the PEP 621 `[project].name` field, which is standard Python project metadata — not "uncoded configuration".
- The other functions in `config.py` (`find_pyproject_toml`, `read_source_roots`, `read_instruction_files`) do read uncoded-specific keys, so the original docstring was accurate before the move.

Either the docstring should broaden (e.g. "Read project metadata and uncoded configuration from pyproject.toml") or it should remain narrow and `read_project_name` should be acknowledged as a small exception via a comment near the function.

Surfaced by the developer agent in the post-merge sweep for PR #45. The drift was introduced by the move in PR #45 and was missed by the in-session coherence review — flagging that gap as part of the report.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #53, opened 2026-04-30
[claude] Maintenance rollup: doc drift, naming, dispatch hardening

alimanfoo:
## Summary

Rolls up several minor maintenance issues into one PR.

### Original-scope issues

- **#42** — `pyproject.toml` description revised to `"Symbol indexes and language-server setup for AI-assisted code navigation"` (was `"Static symbol indexes for AI-assisted code navigation"`). Reflects the project's increased emphasis on language-server setup alongside the static index. General-audience phrasing — names the role rather than the specific tool (Serena).
- **#43** — Added a paragraph to the injected agent instructions (`_SECTION_BODY` in `instruction_files.py`) telling agents to skip Serena's `activate_project` and `check_onboarding_performed` calls. Project is already active by default; `check_onboarding_performed` only gates the `onboarding` flow whose memory-writing surface uncoded deliberately disables.
- **#47** — Reconciled README and `serena_setup.py` module docstring with the actual `SERENA_ALLOWED_TOOLS` / `SERENA_PROJECT_YML` constants (memory tools in README, scope of `excluded_tools`, count of allowlisted tools).
- **#48** — Renamed the integration-level test method `test_falls_back_to_cwd_name_when_no_pyproject` → `test_setup_uses_cwd_name_when_no_pyproject` in `tests/test_serena_setup.py::TestSetup` to disambiguate from the unit-level twin in `tests/test_config.py`.
- **#49** — Widened `config.py` module docstring to acknowledge it now reads PEP 621 project metadata (`read_project_name`) alongside uncoded-specific configuration.
- **#50** — Renamed `_sync_serena_project` → `_write_serena_project_if_absent` so the asymmetry against the two genuinely-syncing helpers (`_sync_mcp_json`, `_sync_claude_settings`) is visible at the call site in `setup()`.
- **#51** — Refactored `cli.main` dispatch via `subparser.set_defaults(action=...)` and `args.action()` to remove the silent fall-through to `_sync` for unrecognised subcommands. Adding a future subcommand without `set_defaults(...)` now raises `AttributeError` rather than silently routing to `_sync`.
- **#52** — Added a one-line comment above `setup()` documenting `root` as a latent test injection seam (the CLI always uses `Path.cwd()`).

### User-directed scope (added in-session)

- Symlinked `CLAUDE.md` to `AGENTS.md`. The two files had been hand-duplicated across many past commits; AGENTS.md is now canonical.
- Deleted `MAINTENANCE.md` (the "Deferred work" parking-lot section was confusing agents — they read it as a parallel issue tracker). Salvaged the PyPI Trusted Publishing release notes into a new `## Releasing` section in README.

### In-session follow-ons (drained depth-first per task)

- Reconciled the hand-written "Approach" section in CLAUDE.md/AGENTS.md with the new pyproject tagline (added a short paragraph noting language-server setup, with a forward pointer to the auto-injected rule section).
- Tightened the Serena skip-guidance wording for precision (`check_onboarding_performed` *gates* the onboarding flow rather than *triggers* memory writing).
- Dropped the now-unread `dest="command"` from `add_subparsers` (clean-up after #51's refactor; spotted by the per-PR reviewer).
- Tightened the AGENTS.md "Commands" code-block comment from "CLAUDE.md section" to "instruction-file section" (now that CLAUDE.md is a symlink; spotted by the per-PR reviewer).

## Verification

Already green at every commit (verified by the lead, audited by the maintainer per-task) and re-runs in CI on this PR — these need no human action:

- `uv run pre-commit run --all-files` — clean (ruff, ruff-format, ty, uncoded sync, plus standard hooks).
- `uv run pytest` — 170 passed.
- `uv run uncoded check` — index up to date.
- `git ls-files -s CLAUDE.md` shows mode `120000`; `readlink CLAUDE.md` is `AGENTS.md`; `cat CLAUDE.md` reads through transparently.
- `TestMainDispatch` covers the refactored CLI dispatch (sync, check, setup, no-subcommand, unknown-subcommand).

Optional human-eyes sanity for the merger:

- Read the regenerated `AGENTS.md` (around line 143) and confirm the new Serena skip-guidance paragraph reads coherently in context.


alimanfoo:
**Recommendation:** Looks good — recommend merge. No blocking concerns. The eight issues are addressed faithfully, the user-directed scope (CLAUDE.md→AGENTS.md symlink, MAINTENANCE.md deletion with release notes salvaged into README) is coherent, the maintenance follow-ons drain depth-first per protocol, all 170 tests pass, `pre-commit run --all-files` is clean, and `uncoded check` reports the index in sync. A few minor consistency items below.

## Non-blocking

1. **README still says "navigation and edit tools" while the module docstring was tightened to "nine tools".** `README.md:172-173` reads `enables the Serena server and allowlists its navigation and edit tools.` The #47 fix correctly updated `serena_setup.py:16-19` to `the nine tools (initial_instructions plus the eight LSP tools — symbol lookup, reference search, and the edit family)`, but the parallel sentence in the README quietly elides `initial_instructions`. The two descriptions sit one click apart on the project's documentation surface, and #47 was specifically about README/code drift on this same constant.

2. **`dest="command"` on `add_subparsers` is now unread.** After the `set_defaults(action=...)` refactor in #51 (`src/uncoded/cli.py:83`), nothing reads `args.command` — dispatch is purely via `args.action()`. The `dest` is harmless but is now dead vocabulary at the very registration site that the refactor was tidying up.

## Nits

1. **Stale `CLAUDE.md section` reference in the canonical instruction file.** `AGENTS.md:41` (the canonical file post-symlink) contains a hand-written code-block comment: `# Generate (or update) the namespace map, stub files, and CLAUDE.md section`. Now that AGENTS.md is canonical and `DEFAULT_INSTRUCTION_FILES` injects into both `CLAUDE.md` and `AGENTS.md`, naming only "CLAUDE.md section" reads as slightly out of date. Something like "CLAUDE.md/AGENTS.md section" or "instruction-file section" would track the new arrangement.

2. **`#52` injection-seam comment lives outside the docstring.** `src/uncoded/serena_setup.py:168`: `# \`root\` is an injection seam for tests; the CLI always uses \`Path.cwd()\`.` is placed immediately above `def setup`, not inside the function's docstring. The docstring already references `root` ("Generate Serena + ty + Claude Code configuration under ``root``."), so a one-line addition there (e.g. "``root`` is an injection seam for tests; the CLI always passes ``None``, falling back to ``Path.cwd()``.") would keep the explanation associated with the symbol when shown by tooling that surfaces docstrings (IDE hover, `help()`, `pydoc`). Issue #52 explicitly listed both options ("comment near the signature" or alternative); the chosen position is fine, just slightly less discoverable.

3. **Asymmetric lambdas in the dispatch.** `src/uncoded/cli.py:92,101,110`: `sync_parser.set_defaults(action=lambda: _sync(check=False))` and `setup_parser.set_defaults(action=lambda: setup())` could both be the bare callable (`_sync` and `setup` respectively) since each has a default-only signature. Only `check_parser.set_defaults(action=lambda: _sync(check=True))` actually needs a wrapper. The current uniform-lambda style is readable as deliberate symmetry; flagging only because it caught the eye.

## Out of scope but noticed

1. **Git symlink on Windows.** `git ls-files -s CLAUDE.md` shows mode `120000`. On Windows clones without `core.symlinks=true`, the working-tree `CLAUDE.md` materialises as a 9-byte text file containing the string `AGENTS.md` rather than a working symlink. Claude Code on Windows would then fail to read the project's instruction file via `CLAUDE.md`. The project's audience may be majority macOS/Linux, but this is a known git-on-Windows wart that the PR introduces; worth knowing.

2. **`uncoded sync` now writes the same file twice via the symlink.** `DEFAULT_INSTRUCTION_FILES = [Path("CLAUDE.md"), Path("AGENTS.md")]` (`src/uncoded/instruction_files.py:17`) — both entries resolve to the same inode in this repo. `sync_instruction_file` is therefore called twice; the second call is a content-identical no-op. Behaviour is correct, but the symlink solution is project-local: downstream uncoded users with both files as separate copies still face the original duplication that prompted the symlink decision.

3. **MAINTENANCE.md deletion removes the project's single-file "Deferred work" channel.** The PR description explains the rationale (agents read it as a parallel issue tracker), and the release-notes content survives in `README.md:210-226`. Worth flagging that the project has lost a checkpoint where humans-and-agents could see "noticed but not fixed" at a glance — GitHub issues are now the only inventory. Intentional; surfacing it because future contributors won't see the prior convention.

4. **README intro paragraph for the language-server angle.** `README.md:12` reads `Additionally, **uncoded** provides some convenience for setting up your coding agent with a language server, so they can reliably find symbol usages and rename, edit, or delete symbols safely.` After the pyproject tagline tightened to `Symbol indexes and language-server setup for AI-assisted code navigation`, this single long sentence at the README's top reads slightly more informal/wordy than its neighbours ("provides some convenience" vs the parallel terseness elsewhere). Pre-existing-shape; flagging only because the new tagline draws the eye to it.

5. **Skip-guidance vs `excluded_tools`.** The new instruction-section paragraph (`src/uncoded/instruction_files.py:114-117`) tells agents to skip both `activate_project` and `check_onboarding_performed`, but `SERENA_PROJECT_YML` already excludes `check_onboarding_performed` (`src/uncoded/serena_setup.py:71`) — so a Serena instance reading that project file shouldn't expose it to begin with. The instruction text is therefore most useful for the unobserved-but-mentioned-in-#43 case where the agent reaches Serena via a different config (or where the exclusion isn't honoured). Coherent, but the relationship between the two defences is not documented anywhere; a future reader auditing "what shields us from these calls?" has to assemble it from two files.

PR #67, opened 2026-05-01
[claude] Post-#53 maintenance rollup

### LINKED ISSUE #60: [claude] setup() root parameter is dormant; tests use os.chdir without restoration
The `setup(root: Path | None = None)` function in `src/uncoded/serena_setup.py` exposes a `root` parameter that no current caller passes:

- `cli.main` invokes `setup()` with no args; the CLI always falls back to `Path.cwd()`.
- Tests in `tests/test_serena_setup.py::TestSetup` use `os.chdir(tmp_path)` and call `setup()` argless.

The parameter was annotated in PR #53 as "an injection seam for tests; the CLI always uses Path.cwd()" — but tests don't actually exercise the seam.

Two coupled concerns:

1. The `root` parameter is documented as a test seam but is genuinely dormant — no caller exercises it.
2. Tests using `os.chdir(tmp_path)` rely on pytest not restoring cwd between tests. A failure mid-test leaves the process in a no-longer-existent `tmp_path`; subsequent tests in unrelated modules can inherit the wrong cwd silently. Hidden flakiness risk.

Possible directions:
- Refactor tests to pass `root=tmp_path` instead of `os.chdir`. Justifies the seam, eliminates the cwd-restore-on-failure risk.
- Remove the `root` parameter entirely. Admits the seam was never used; tests stay on the `os.chdir` pattern (with a fixture that restores cwd if flakiness becomes a concern).

Either is coherent; current state has the costs of both.

Sources: developer agent observation during PR #53 task #6, and developer agent post-merge sweep.

### LINKED ISSUE #61: [claude] _STATUS_VERB dict in serena_setup.py is an unenforced string protocol
`src/uncoded/serena_setup.py::_STATUS_VERB` is a dict keyed by `"wrote"` / `"updated"` / `"unchanged"`. Three helpers (`_sync_mcp_json`, `_write_serena_project_if_absent`, `_sync_claude_settings`) return strings expected to match these keys, and `setup()` indexes into `_STATUS_VERB` with each return value when printing per-file status.

The helpers' return type is bare `str`, so a typo (e.g. `return "writen"`) escapes ty's notice and surfaces as a runtime `KeyError` in `setup()`'s status-print loop.

Using `typing.Literal["wrote", "updated", "unchanged"]` (or a `StrEnum`) for both the helper return type and the dict keys would let ty catch typos at type-check time.

Sources: developer agent post-merge sweep on PR #53.

### LINKED ISSUE #62: [claude] tests/test_serena_setup.py asserts only some args of MCP_SERVER_SERENA
`tests/test_serena_setup.py::test_mcp_json_is_valid_and_pins_version` checks several args of `MCP_SERVER_SERENA["args"]` via `in serena["args"]` membership, but several args are not asserted at all:

- `--open-web-dashboard`
- The `"false"` literal that pairs with it
- `--transport`
- `stdio`
- `--from`

A drop, typo, or re-ordering in any of those tokens wouldn't surface as a test failure — it would just be a quietly-broken Serena launch.

Possible directions:
- Add an `EXPECTED_MCP_ARGS` constant in the test file (mirroring `EXPECTED_EXCLUDED_TOOLS`) and assert the full args list with set-equality or order-equality.
- Add individual `in` assertions for each currently-unchecked token.

The set-equality variant is more durable and matches the existing pattern for `EXPECTED_EXCLUDED_TOOLS`.

Sources: developer agent post-merge sweep on PR #53.

### LINKED ISSUE #63: [claude] uncoded's index does not include module docstrings
uncoded's static index does not capture module docstrings:

- `src/uncoded/stubs.py::extract_stub` captures function/class first-sentence docstrings, but not module docstrings.
- `src/uncoded/namespace_map.py::build_map` does not carry docstrings at all.

The `config.py` module-docstring drift fixed in #49 was therefore invisible to uncoded's own coherence-review skill — agents reading the index could not see that the docstring's claim ("Read uncoded configuration from pyproject.toml") had drifted from the file's actual contents (now also reading PEP 621 project metadata via `read_project_name`).

This touches uncoded's own value proposition. The README frames the problem as "agents fill gaps from pretraining rather than reading the actual code"; module docstrings are exactly the kind of orienting prose that prevents that, and the project's static index can't help an agent catch drift in its own.

Possible directions:
- Add module docstrings to the per-file stub (`.uncoded/stubs/*.pyi`). The stub already captures all other first-sentence docstrings in the file; the module docstring would land naturally at the top.
- Add module docstrings to the namespace map. Less natural — the namespace map is a symbol skeleton and currently carries no docstrings.
- The stub option is mildly preferred because it doesn't expand the namespace map's scope.

Sources: developer agent post-merge sweep on PR #53.

### LINKED ISSUE #64: [claude] _write_serena_project_if_absent doesn't escape project_name in YAML template
`_write_serena_project_if_absent` in `src/uncoded/serena_setup.py` interpolates `project_name` into the `SERENA_PROJECT_YML` template via `str.format(project_name=...)`.

A project name containing:

- `"` (quotation marks) → produces malformed YAML
- `{` or `}` → format-spec collision; `str.format` raises `KeyError`/`IndexError` or silently misinterprets
- `\` (backslashes) → unintended escape sequences
- YAML special characters (`:`, `-`, `&`, `*`, `!`, `|`, `>`) → ambiguous YAML

`read_project_name` (in `src/uncoded/config.py`) accepts whatever PEP 621 `[project].name` happens to be, which is more permissive than the YAML template tolerates.

Edge case — most names are simple identifiers — but a sharp edge with no test exercising it.

Possible directions:
- Switch to `yaml.dump`-based construction for the YAML, with the `project_name` field properly escaped by the YAML library.
- Validate `project_name` against a simple-identifier regex at the boundary and reject names that wouldn't round-trip safely.
- Escape characters before interpolation (less robust than yaml.dump).

Sources: developer agent post-merge sweep on PR #53.

### LINKED ISSUE #65: [claude] uncoded sync prints "Updated CLAUDE.md" only when CLAUDE.md is a symlink to AGENTS.md
`sync_file` in `src/uncoded/sync.py` writes via `Path.write_text(content)`, which follows symlinks. When `uncoded sync` iterates `DEFAULT_INSTRUCTION_FILES = [Path("CLAUDE.md"), Path("AGENTS.md")]` in a repo where CLAUDE.md is a symlink to AGENTS.md (like this one):

- Pass 1 (CLAUDE.md, the symlink) reads through to AGENTS.md, generates the section, writes through the symlink. If content differs, prints `Updated CLAUDE.md`.
- Pass 2 (AGENTS.md) reads, finds content already correct (just written via the symlink in pass 1), returns without printing.

Result: a developer who hand-edits the navigation section and runs `uv run uncoded sync` sees only `Updated CLAUDE.md` in the output. AGENTS.md never appears, even though that's the file actually on disk.

Semantically correct (one inode, one write) but cosmetically asymmetric. Could prompt "was AGENTS.md updated too?" from a contributor unfamiliar with the symlink.

Possible directions:
- Detect the symlink case in the iteration logic and print something like `Updated CLAUDE.md (-> AGENTS.md)`.
- Dedupe the iteration by `path.resolve()` so each inode is processed once; print the canonical (resolved) path.
- Leave as-is and add a brief explanatory comment in `sync_instruction_file` for future readers.

Sources: maintainer agent post-merge sweep on PR #53.

### LINKED ISSUE #66: [claude] Miscellaneous minor documentation improvements
A grouped collection of low-priority documentation polish items surfaced during PR #53 review and post-merge sweeps. Each is small in isolation; bundled here to keep the issue tracker uncluttered.

Each item below is independently addressable. A future PR could pick any subset.

## 1. Windows symlink behaviour for CLAUDE.md → AGENTS.md

After PR #53, `CLAUDE.md` is a `120000` (symlink) entry in git pointing at `AGENTS.md`. On Windows clones without `core.symlinks=true`, the working-tree `CLAUDE.md` materialises as a 9-byte text file containing the string `AGENTS.md` rather than a working symlink. Tools reading the project's instruction file via the `CLAUDE.md` path would then fail.

Project audience is likely majority macOS/Linux (CI is Linux-only), but the pathology is real. Worth a short note in README's contributor section, or a project-direction decision.

Files: `CLAUDE.md`, `AGENTS.md`, `.github/workflows/ci.yml`, `README.md`.

## 2. README intro paragraph wordiness

`README.md:12` reads:

> Additionally, **uncoded** provides some convenience for setting up your coding agent with a language server, so they can reliably find symbol usages and rename, edit, or delete symbols safely.

After the pyproject tagline tightened in #42 to `"Symbol indexes and language-server setup for AI-assisted code navigation"`, the README intro reads slightly more informal/wordy than its surroundings ("provides some convenience" vs. the parallel terseness elsewhere). Pre-existing prose; the new tagline draws the eye to it.

Files: `README.md:12`.

## 3. Defense-in-depth (skip-guidance + excluded_tools) is not documented

The new skip-guidance paragraph in `_SECTION_BODY` (added in #43) and `SERENA_PROJECT_YML.excluded_tools` (in `serena_setup.py`) form two complementary defensive layers:

- Server-side narrowing in `.serena/project.yml` hides `check_onboarding_performed` and `onboarding` from Serena's tool surface entirely.
- Agent-side guidance tells the agent to skip those calls plus `activate_project`.

The two layers reinforce rather than conflict (defense in depth: the skip-guidance is also a real safeguard for users who run `uncoded sync` without `uncoded setup`, where the exclusions aren't yet applied). But the relationship isn't documented anywhere; a future reader auditing "what shields us from these calls?" has to assemble it from two files.

A brief design note — either in `serena_setup.py`'s module docstring, or `instruction_files.py`'s, or a top-level architecture doc — would help.

Files: `src/uncoded/serena_setup.py:9-26` (module docstring), `src/uncoded/instruction_files.py` (module docstring), or a new top-level doc.

## 4. instruction_files.py library docstring doesn't acknowledge the symlink pattern

`src/uncoded/instruction_files.py` describes the library's contract for projects that want to support both AGENTS.md and CLAUDE.md as instruction files. The dogfooded uncoded repo now uses a symlink (CLAUDE.md → AGENTS.md) as its way of satisfying that contract, but the library's docstring describes only the two-files case. Worth acknowledging the symlink pattern as a supported configuration.

Files: `src/uncoded/instruction_files.py:1` (module docstring).

## 5. "Pure LSP operations" terminology drift in serena_setup.py module docstring

`src/uncoded/serena_setup.py:11` says `.serena/project.yml` "narrows Serena's surface to pure LSP operations". After the docstring update in #47, lines 17-19 now explicitly carve out `initial_instructions` as a non-LSP bootstrap tool included in the allowlist. The line-11 "pure LSP operations" phrasing is therefore slightly loose — the YAML's exclusions of memory/onboarding/dashboard/shell-exec don't actually narrow the surface to *only* LSP; the final narrowing to LSP-plus-bootstrap happens in `.claude/settings.json`. Pre-existing wording; #47 made it slightly more visible.

Files: `src/uncoded/serena_setup.py:11`.

## 6. "dispatch rule" terminological seam in the Approach paragraph

The hand-written Approach paragraph at `AGENTS.md:32` (added in PR #53 follow-on) ends:

> See "How to read and edit code in this codebase" below for the dispatch rule.

But the auto-injected target section uses different terms:

- `AGENTS.md:56`: heading `### The rule` — no "dispatch" qualifier.
- `AGENTS.md:158`: closing line "The dispatch test is the search term: ..." — uses "dispatch test".

So the cross-reference synthesises a phrase ("dispatch rule") not used at the target. Heading-name pointer makes navigation unambiguous, so this isn't a navigation bug — but the rule / dispatch rule / dispatch test drift across three surfaces caught the eye.

Files: `AGENTS.md:32`, `AGENTS.md:56`, `AGENTS.md:158`, source: `src/uncoded/instruction_files.py::_SECTION_BODY`.

## 7. AGENTS.md Commands section enumerates only 2 of 3 uncoded subcommands

The hand-written Commands section (`AGENTS.md:34-46`) lists `uv run uncoded sync` and `uv run pytest`, but the project has three `uncoded` subcommands: `sync`, `check`, and `setup`. The section reads as a quick-reference for contributors, but `check` (the CI gate) and `setup` (the Serena bootstrap) are absent. Adjacent to the Commands-comment tightening in PR #53, which sharpened a comment within an otherwise still-incomplete command list.

Files: `AGENTS.md:34-46`.

---

Sources: PR #53 reviewer review, maintainer per-task and post-merge audits, developer post-merge sweep.

alimanfoo:
A maintenance batch addressing the seven sweep issues that fell out of PR #53. One issue per task, one task per commit, with maintainer follow-on chains drained depth-first as the protocol prescribes.

## What changes, by issue

**#62** — Assert the full `MCP_SERVER_SERENA` args list in tests rather than checking partial membership; align the dogfooding test to the same shape; rename it to `test_repo_mcp_json_matches_template_contract` for consistency with its sibling.

**#61** — Type `_STATUS_VERB` keys as a `Literal["wrote", "updated", "unchanged"]` alias so a typo in any of the three sync helpers' return values becomes a ty error rather than a runtime `KeyError` when `setup()` formats the per-file status line.

**#64** — Switch `.serena/project.yml` writing from string-template interpolation to `yaml.safe_dump`, so YAML-special characters in `project_name` are escaped instead of producing malformed YAML. The constant `SERENA_PROJECT_YML` (formerly a YAML string template) is now `SERENA_PROJECT_FIELDS` (a Python dict that's serialised at write time).

**#60** — Refactor `TestSetup` to pass `root=tmp_path` to `setup()` explicitly rather than relying on `os.chdir`. Dropping the chdir required mocking `read_project_name` in the test helper, because its cwd-walk would otherwise reach uncoded's own `pyproject.toml`. A separate test exercises the pyproject path through `setup()` end-to-end so the integration coverage isn't lost.

**#65** — Dedupe the instruction-file iteration in `cli._sync` by resolved inode. When `CLAUDE.md` is a symlink to `AGENTS.md`, sync now processes the file once under the canonical name rather than printing `Updated CLAUDE.md` and going silent on `AGENTS.md`.

**#63** — Extend the stub generator to capture module docstrings (first-sentence convention, matching how class and function docstrings are already handled). Every existing stub regenerated to gain its module-docstring header. The drift fixed in #49 — a stale module docstring whose claim no longer matched the code — would now be visible to agents reading the index.

**#66** — Seven small documentation polish items: a Windows-symlink note for contributors, a tightened README intro, a defense-in-depth design note in `serena_setup.py`, acknowledgement of the symlink configuration in `instruction_files.py`, a more accurate rewording of "pure LSP operations", terminology consistency on "dispatch rule" across the three named locations, and the `check`/`setup` subcommands added to the AGENTS.md Commands block. A follow-on closed a misstatement: `activate_project` was not actually excluded at the YAML layer, so it's been added to `SERENA_PROJECT_FIELDS["excluded_tools"]` (and the test canonical, and this repo's own `.serena/project.yml`) to match the documented defense-in-depth claim.

Closes #60
Closes #61
Closes #62
Closes #63
Closes #64
Closes #65
Closes #66


alimanfoo:
## Reviewer first pass

**Recommendation:** approve subject to nits — twelve commits do what each issue asked for, and the dogfood/test surface tightens around the changes; only small cosmetic things to flag.

## Nits

1. The new dedupe loop in `src/uncoded/cli.py:55` calls its set `seen_inodes: set[Path]` and the comment says "Dedupe configured instruction paths by resolved inode," but the set actually holds resolved `Path` objects and the comparison is path-equality. The comment then switches to "Resolving collapses both aliases to the same canonical path," which is the accurate description. The behaviour is correct (resolved paths collapse symlinks), but `seen_resolved` (or `seen_canonical`) and a single consistent term would read more cleanly than mixing "inode" and "canonical path" two lines apart.

2. Same block, the comment at `src/uncoded/cli.py:54` says "we then render relative to cwd for the user-facing line." The `relative_to(Path.cwd())` call is wrapped in a `try/except ValueError` that falls back to the absolute resolved path. So for an instruction file configured outside cwd (e.g. `../docs/foo.md`), the user-facing line will be an absolute path, not a relative one. Functionally fine; the comment just slightly overstates what the rendering does.

3. `src/uncoded/__init__.py` gains a `"""Uncoded."""` docstring purely so module-docstring extraction (#63) has something to pick up at the package root. The stub now reproduces it. It carries no information a reader doesn't get from the package name itself — consider dropping it (the stub generator handles missing module docstrings cleanly via the test added in `test_module_no_docstring`).

4. `tests/test_cli.py:90` filters captured stdout with `endswith("AGENTS.md") or endswith("CLAUDE.md")` and asserts the filtered list equals `["Updated AGENTS.md"]`. That catches the regression the test is named for, but it would also silently swallow an unexpected extra line that didn't end in either name. Asserting the full instruction-related output rather than a filtered slice would make the test stricter without much extra code; arguably overkill for a deduplication sanity check, hence "nit".

5. `tests/test_serena_setup.py::test_setup_uses_cwd_name_when_no_pyproject` and the new `test_setup_reads_name_from_pyproject` both `monkeypatch.chdir(tmp_path)` *and* pass `root=tmp_path`. With cwd and root pointing at the same directory, a regression to `setup()` ignoring `root` and using `Path.cwd()` directly would still pass these two tests. The `_run` helper's comment correctly notes that "the seam is meaningfully tested only when cwd ≠ root," and the mocked tests via `_run` do exercise that case — these two end-to-end tests are about name resolution rather than the seam, so the property is covered overall, just not within these particular tests.

## Out of scope but noticed

1. The new test for the symlink-dedup case covers apply mode only (`cli._sync()`). Check mode (`cli._sync(check=True)`) runs the same dedup loop, but symmetry isn't asserted. Probably fine, since both modes share the iteration code, but flagging in case the lead wants it on the post-merge list.

2. `_Status` is the project's first PEP 695 `type` alias and lands in `.uncoded/namespace.yaml` indented at the same level as ordinary constants, with no marker that it's a type alias. The stub generator already tracks `is_type_alias` for PEP 695 aliases (via `StubAssignment`), but the namespace map's flat listing doesn't reflect that. Pre-existing limitation, not introduced by this PR — just newly visible because this is the first such alias in the codebase.

PR #76, opened 2026-05-01
[claude] Post-#67 follow-on cleanup

### LINKED ISSUE #68: [claude] setup(root=X) splits cwd from output dir but read_project_name still walks from cwd
A caller of `setup(root=some_dir)` from a different cwd writes config files into `some_dir` while reading the project name from whatever `pyproject.toml` cwd happens to walk to. The two halves of "what project are we configuring" can silently disagree.

`setup()` in `src/uncoded/serena_setup.py` honours `root` for output paths. `read_project_name()` in `src/uncoded/config.py` walks up from `Path.cwd()` looking for `pyproject.toml`. PR #67's #60 fix made this asymmetry visible by exercising the seam in tests — `TestSetup` now passes `root=tmp_path` with cwd unchanged, which forced `read_project_name` to be mocked because otherwise its cwd-walk would reach uncoded's own `pyproject.toml`.

Possible directions:
- Thread `root` into `read_project_name` so the project-name lookup honours the same directory as the output.
- Add an explicit `project_name` parameter to `setup()` so the caller resolves the name however makes sense for them.
- Document the asymmetry as intentional (the CLI never crosses cwd vs. output, so it doesn't bite in normal use).

Sources: developer agent observation during PR #67 task #4 implementation; surfaced again in the post-merge sweep.

### LINKED ISSUE #69: [claude] Symlink-dedup asserted only for cli._sync apply mode, not check mode
`tests/test_cli.py::TestSyncApplyMode::test_dedupes_when_claude_md_is_symlink_to_agents_md` exercises `cli._sync()` (apply mode) and asserts the canonical-name output. Check mode (`cli._sync(check=True)`) runs the same dedup loop but the symmetry isn't asserted. A regression that broke dedup only in one path — for instance, future code that branches on `check` and re-introduces a separate iteration — would slip past.

Both modes share the iteration code today, so the gap is small in practice. PR #67 added the apply-mode test as the minimum to prove the fix; the parallel for check mode wasn't added.

Possible direction: add a sibling test in `TestSyncCheckMode` that mirrors the apply-mode shape (CLAUDE.md as symlink to AGENTS.md, run `cli._sync(check=True)`, assert exactly one instruction-file line and the canonical name). Cheap and structurally parallel to the existing test.

Sources: reviewer agent on PR #67.

### LINKED ISSUE #71: [claude] tests/test_cli.py uses bare os.chdir; same hygiene gap that #60 fixed in test_serena_setup.py
`tests/test_cli.py::_init_repo` calls bare `os.chdir(tmp_path)`, and roughly ten test methods across `TestSyncApplyMode`, `TestSyncCheckMode`, and the error-path tests rely on it. On test failure, cwd leaks; subsequent tests in unrelated modules can inherit the wrong cwd silently. This is the same flakiness pattern that issue #60 fixed for `tests/test_serena_setup.py::TestSetup` in PR #67.

Pre-existing in `test_cli.py`. PR #67's scope was `test_serena_setup.py` specifically, so the parallel pattern wasn't pulled in.

Possible direction: refactor `_init_repo` and its callers to use `monkeypatch.chdir(tmp_path)` (auto-restored on teardown including failure). For tests that exercise behaviour where cwd vs. an explicit working directory matter, also pass an explicit working directory through the API where one is supported.

Sources: developer + maintainer agents post-merge sweep on PR #67.

### LINKED ISSUE #72: [claude] _first_sentence truncates docstrings starting with 'e.g.', 'i.e.', 'Mr.', etc.
The first-sentence extractor in `src/uncoded/stubs.py::_first_sentence` treats any `.` followed by whitespace as a sentence boundary. Docstrings beginning with a common abbreviation get truncated to the abbreviation alone:

- "e.g. parse a YAML file..." → excerpt becomes "e.g."
- "i.e. ..." → excerpt becomes "i.e."
- "Mr. ..." / "Dr. ..." / "U.S. ..." — same shape.

Pre-existing, but PR #67's issue #63 fix (extracting module docstrings) widened the blast radius. Abbreviations show up in prose intros more often than in single-line function summaries, so module-docstring extracts are more exposed.

Possible directions:
- Match `.` followed by whitespace AND a capital letter (catches most natural sentence boundaries; might still miss "U.S. is a country..." style).
- Whitelist a small set of common abbreviations to skip past.
- Use a proper sentence tokeniser (overkill for the small input).

The capital-letter heuristic is smallest and likely sufficient.

Sources: developer agent post-merge sweep on PR #67.

### LINKED ISSUE #74: [claude] iter_source_files silently skips files that fail to parse
`src/uncoded/extract.py::iter_source_files` swallows `SyntaxError` (per `tests/test_extract.py::test_skips_syntax_errors`) so a temporarily-broken Python module disappears from the namespace map and stubs subtree with no signal. Effects:

- An agent reading `.uncoded/namespace.yaml` won't know the file exists or that there's something to look for.
- A CI freshness check (`uv run uncoded check`) produces stable output for the broken state, so it won't flag the divergence either.

The skip-on-syntax-error policy itself is fine — the alternative (failing the build) would block normal in-progress edits. The problem is silence: there's currently no breadcrumb that anything was skipped.

Possible direction: emit a stderr line when skipping ("warning: skipping {path}: SyntaxError at line {n}: {message}"). Cheap, doesn't change the contract, gives the developer a chance to notice.

Sources: developer agent post-merge sweep on PR #67.

### LINKED ISSUE #75: README does not mention uncoded setup
To wire up the language server requires running `uncoded setup`. However, this isn't currently mentioned anywhere in the README. We should add something, at the natural place for a new user wanting to install uncoded and then set it up with their repo, following through the README.

alimanfoo:
PR #67's post-merge sweep surfaced six follow-on issues. This branch addresses them as a rollup, one commit per issue. Each commit's message names its issue.

- **#71 — Use `monkeypatch.chdir` in `tests/test_cli.py`.** Same hygiene gap #60 fixed in `test_serena_setup.py` in PR #67. Bare `os.chdir` could leak cwd to subsequent tests on failure; `monkeypatch.chdir` restores cwd on teardown including failure.
- **#69 — Test symlink dedup in check mode.** `cli._sync(check=True)` shares the dedup loop with apply mode but the symmetry wasn't asserted; structural insurance against a future regression that branches on `check`.
- **#72 — Stop truncating docstrings at abbreviations.** `_first_sentence` in `src/uncoded/stubs.py` now requires a capital after the period for the boundary to fire, so docstrings beginning with `e.g.`, `i.e.`, `U.S.` etc. aren't truncated to the abbreviation alone. The `Mr. Smith arrived.` shape is a known limitation of this heuristic and stays open as residual.
- **#74 — Warn on syntax-error skip in `iter_source_files`.** Files that fail to parse are still skipped (as before) but now leave a stderr breadcrumb (`warning: skipping {path}: SyntaxError at line {n}: {msg}`) so a stale-stub-from-broken-source has a paper trail. Skip is centralised into `iter_source_files`; `walk_source` and `_generate_stubs` shed their redundant try/excepts.
- **#68 — Thread `setup` root into config readers.** `setup(root=X)` honoured `root` for output paths but `read_project_name` walked from cwd, so the two halves of "what project are we configuring" could disagree. All four `config.py` readers now take a required `start: Path`. `setup` passes `start=root`; `cli._sync` binds `cwd = Path.cwd()` and threads it. `tests/test_serena_setup.py` no longer needs to mock `read_project_name` — the seam is exercised end-to-end via a real `pyproject.toml` at `tmp_path` with cwd deliberately unaligned.
- **#75 — Surface `uncoded setup` in the README.** A new "Set up the language server" section between "Use" and "Keep it current with pre-commit" introduces the command and links forward to the existing detailed section, so a top-down reader doesn't miss it.

Closes #68
Closes #69
Closes #71
Closes #72
Closes #74
Closes #75


alimanfoo:
**Recommendation:** approve subject to a couple of nits

## Non-blocking
1. **The "Mr. Smith arrived." residual isn't named in the source comment.** The PR description and #72 both acknowledge that the new heuristic still truncates docstrings beginning with a title+capital-name pair (`Mr. Smith…`, `Dr. Jones…`) to just the abbreviation, but the inline comment in `src/uncoded/stubs.py:81-87` only lists the cases the heuristic now *handles* (`e.g.`, `i.e.`, `U.S.`). Once #72 closes, that limitation is only recorded in the issue history. Worth a one-liner in the comment so a future maintainer extending `_first_sentence` knows which input shape is deliberately out of scope.

## Nits
1. **`_run` helper interpolates `name` into TOML without escaping.** `tests/test_serena_setup.py:54` writes `f'[project]\nname = "{name}"\n'`. Today's callers (`"my-app"`, `"weird: name [with] # chars"`) all happen to be safe — none of them contain `"` or `\`, the only characters TOML basic strings need to escape. A future caller passing either character would silently produce invalid TOML and a confusing test failure.

2. **`uncoded setup` vs `uv run uncoded setup` in the README.** The new section at `README.md:71-73` uses bare `uncoded setup`, while the more detailed "Using uncoded with a language server" section a few headings below uses `uv run uncoded setup`. The new section's style matches the immediately-preceding `uncoded sync` block, so this is probably intentional, but the two stanzas now disagree.

## Out of scope but noticed
1. Bare `os.chdir(tmp_path)` is still alive in `tests/test_skill.py` (10 call sites across `TestSyncSkill`) and in the `_setup` helpers of `tests/test_stubs.py::TestBuildStubs` and `TestBuildStubsCheckMode`. Same hygiene gap that #71 fixes in `tests/test_cli.py`; cwd leakage on test failure can silently affect downstream tests. Pre-existing and out of this PR's stated scope.

2. `iter_source_files` is invoked twice per `uncoded sync` run for each source root — once via `walk_source` at `src/uncoded/cli.py:41` for the namespace map, and again via `build_stubs` → `_generate_stubs` for the stubs subtree at `src/uncoded/cli.py:47`. With the #74 warning now centralised inside `iter_source_files`, a single broken file will produce two identical `warning: skipping …` lines per sync run. Not a bug in the fix itself — a side effect of the call graph the centralisation now exposes.

Issue #54, opened 2026-04-30
[claude] setup(root=...) is a dormant API; tests mutate global cwd instead


alimanfoo:
`src/uncoded/serena_setup.py::setup` exposes a `root: Path | None = None` parameter that's not currently exercised by any caller:

- The CLI entry point (`src/uncoded/cli.py::main`) calls `setup()` with no args.
- The test fixtures `tests/test_cli.py::_init_repo` and `tests/test_serena_setup.py::TestSetup._run` both use `os.chdir(tmp_path)` and call `setup()` argless, relying on the `Path.cwd()` fallback inside `setup()`.

The parameter is plumbed through `setup()`'s body (default-resolution, path construction, relative-path output) but never receives a non-`None` value in production or tests. A comment was added in PR #53 acknowledging it as a "latent injection seam for tests"; this issue is for the underlying decision: keep, remove, or actually use as the test seam.

The cwd-mutating test pattern is also a hidden-flakiness risk in its own right: pytest doesn't auto-restore `cwd` between tests, so a failure mid-test inside one of these classes leaves the process in a no-longer-existent `tmp_path` for whatever runs next. Subsequent tests in unrelated modules can inherit the wrong cwd silently.

Switching the tests to `setup(root=tmp_path)` would both exercise the seam and remove the global mutation, justifying the parameter's existence.

Surfaced by the developer agent during task #6 implementation and again in the post-merge sweep for PR #53.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

alimanfoo:
Duplicate of #60. Closing — first batch of `gh issue create` calls reported an internal tool error so I retried, but the originals had silently succeeded. #60 is the canonical version of this concern.

PR #83, opened 2026-05-01
[claude] cli._sync working-directory contract + post-#76 cleanup

### LINKED ISSUE #77: [claude] cli._sync still binds cwd implicitly after #68's reader-level fix
PR #76 closed #68 by threading `start: Path` through every config reader, so `setup(root=X)` and `cli._sync` both spell their working directory explicitly when calling `read_project_name`, `read_source_roots`, etc. The contract gap that #68 fixed at the reader frame still sits one frame up: `cli._sync` reads `Path.cwd()` once and threads it to the readers, but several other working-directory dependencies inside the function are not threaded.

Three angles surfaced in PR #76's post-merge sweep, all citing `cli._sync`:

1. `[r.resolve() for r in read_source_roots(cwd)]` at `src/uncoded/cli.py:29` resolves source-root paths against cwd, not against the directory `find_pyproject_toml(cwd)` actually located. Running `uncoded sync` from a subdirectory under the project root finds the parent's `pyproject.toml` correctly, but `Path("src").resolve()` lands inside the subdir.
2. `walk_source(root)` at `src/uncoded/cli.py:41` and `build_stubs(root, check=check)` at `src/uncoded/cli.py:47` don't take a `base` argument, so they call `iter_source_files(..., base=None)` which defaults internally to `Path.cwd()`. Two of the three downstream sites in `_sync` use the explicitly-bound `cwd`; these three silently re-fetch it.
3. `cli._sync()` itself has no `root=` parameter analogous to `setup(root=...)`. The `tests/test_cli.py` suite still relies on `_init_repo`'s `monkeypatch.chdir` to set up its world; the asymmetry between the two CLI entries is now visible — one is a function with a clean test boundary, the other still requires a process-state mutation to test.

Prior chips on this surface:

- #54: setup(root=...) is a dormant API; tests mutate global cwd instead.
- #60: setup() root parameter is dormant; tests use os.chdir without restoration.
- #68: setup(root=X) splits cwd from output dir but read_project_name still walks from cwd.

Each chip pushed the seam one frame deeper. The contract question worth asking explicitly: what does `cli._sync` promise its caller about working directory, and where does the implementation rely on that promise being implicit?

Possible direction: give `_sync` a `root: Path | None = None` parameter analogous to `setup(root=...)`, defaulting to `Path.cwd()` at the CLI boundary; thread `root` through the `Path.resolve()` of source roots; give `walk_source` and `build_stubs` (and `iter_source_files`'s callers) explicit `base` parameters and pass `root` through. Side-effect: `_sync` becomes testable against arbitrary roots, and the remaining bare `os.chdir` in `tests/test_skill.py` / `tests/test_stubs.py` (a separate hygiene chip) becomes naturally absorbable, just as #68's fix absorbed `tests/test_config.py`.

### LINKED ISSUE #78: [claude] tests/test_skill.py and tests/test_stubs.py still use bare os.chdir
`tests/test_skill.py::TestSyncSkill` (10 call sites) and the `_setup` helpers in `tests/test_stubs.py::TestBuildStubs` and `TestBuildStubsCheckMode` call bare `os.chdir(tmp_path)` to set up the test working directory. On test failure, cwd doesn't get restored; subsequent unrelated tests can inherit the wrong cwd silently. Same hygiene gap that #60 fixed in `tests/test_serena_setup.py` and #71 fixed in `tests/test_cli.py`.

Same fix shape as those chips: switch to `monkeypatch.chdir(tmp_path)`, which restores cwd on teardown including failure.

Prior in-kind fixes:

- #60: setup() root parameter is dormant; tests use os.chdir without restoration.
- #71: tests/test_cli.py uses bare os.chdir; same hygiene gap that #60 fixed.

Note: this chip is also a downstream symptom of the `cli._sync` working-directory contract gap (see #77). If `_sync` and `build_stubs` gain explicit working-directory parameters, several of these `os.chdir` calls fall away naturally — `tests/test_config.py`'s `os.chdir` was absorbed in exactly that way by #68's fix. A purely mechanical sweep is also viable independent of that broader change.

### LINKED ISSUE #79: [claude] iter_source_files emits duplicate skip warnings per uncoded sync run
A single source file with a `SyntaxError` produces two identical `warning: skipping {path}: SyntaxError at line {n}: {msg}` lines on stderr per `uncoded sync` invocation, not one.

Cause: `cli._sync` invokes `iter_source_files` twice for each source root — once via `walk_source(root)` at `src/uncoded/cli.py:41` (for the namespace map) and once via `build_stubs(root, check=check)` → `_generate_stubs` at `src/uncoded/cli.py:47` (for the stubs subtree). #74's centralisation of the syntax-error skip into `iter_source_files` means each invocation re-reads, re-parses, and re-warns about the same broken file.

Side-effect of #74's fix; not present before. The pre-#74 behaviour silently skipped the file twice; now it noisily warns twice.

Possible direction: have `iter_source_files` yield `(source, rel_path, ast.AST)` tuples so downstream consumers don't re-parse, and have `walk_source` and `build_stubs` share results from a single `iter_source_files` invocation — or have `cli._sync` cache the iterator's output and feed both consumers from one read. Either way the warning becomes once-per-broken-file-per-sync.

### LINKED ISSUE #81: [claude] _first_sentence's fallback branch is now narrowly reachable; comment doesn't say what it's for
After PR #76's #72 fix added the `" Z"` sentinel to `src/uncoded/stubs.py::_first_sentence`, the line-based fallback at the end of the function (`return text.split("\n")[0].strip()`) is reachable only for docstrings that contain no period at all (e.g. `"Hello world"`, `"What now?"`).

The function's inline comment block now narrates the regex+sentinel construction in detail but says nothing about what the fallback is for. A reader working through the function fresh has to reason backwards from the regex to figure out which inputs reach the final return.

Pre-PR-#76 the fallback fired for any docstring without `.` followed by whitespace — a much wider input class. The narrowing was a deliberate side-effect of the sentinel, but the comment's framing is still set to the pre-sentinel reachability.

Possible direction: add a one-line comment naming the no-period case as the fallback's remaining home. Or, if the no-period case is unimportant enough to handle the same way as the matched case (e.g. just always return the first line), simplify the function down to that and drop the fallback branch entirely.

alimanfoo:
Four follow-on issues from the #76 post-merge sweep. The biggest is making `cli._sync`'s working-directory contract for *reads* explicit; the rest are smaller cleanups in adjacent code.

## `cli._sync` working-directory contract

`_sync` now takes a `root` parameter (defaulting to `Path.cwd()`) and threads the located project root through every project-relative *input*: source-root resolution, instruction-file lookup, and the rel-paths rendered into the namespace map and stubs. Paths configured in `pyproject.toml` now resolve against the directory containing the toml rather than against whatever cwd happens to be, so running `uncoded sync` from a subdirectory now finds the right sources.

Output paths (`.uncoded/namespace.yaml`, `.uncoded/stubs/...`, instruction-file writes, the skill files) still resolve relative to cwd. Bringing those onto the same anchor is a separate, larger concern that touches `sync_skill`'s home-directory model — tracked outside this PR.

## One skip warning per broken file per sync

When a source file fails to parse, `iter_source_files` warns once on stderr and skips it. Before this PR, the namespace-map and stubs pipelines each ran `iter_source_files` separately, producing two identical warnings per broken file per `uncoded sync` invocation. `_sync` now drives a single iteration per source root and feeds both pipelines from that pass; new pure extraction functions (`extract_modules`, `_generate_stubs`) handle the work without re-walking.

## Test hygiene

`tests/test_skill.py`'s 10 bare `os.chdir(tmp_path)` calls become `monkeypatch.chdir(tmp_path)`, so cwd is restored on test teardown including failure. The matching calls in `tests/test_stubs.py` are absorbed by `build_stubs` taking an explicit `base=` and removed entirely.

## `_first_sentence` simplification

The docstring-excerpt extractor in `src/uncoded/stubs.py` collapses to a single regex with two alternatives (sentence boundary or first-line boundary), removing the explicit fallback branch.

Closes #77
Closes #78
Closes #79
Closes #81

### DIFF
--- a/src/uncoded/cli.py
+++ b/src/uncoded/cli.py
@@ -4,7 +4,11 @@ import argparse
 import sys
 from pathlib import Path
 
-from uncoded.config import read_instruction_files, read_source_roots
+from uncoded.config import (
+    find_pyproject_toml,
+    read_instruction_files,
+    read_source_roots,
+)
 from uncoded.extract import walk_source
 from uncoded.instruction_files import sync_instruction_file
 from uncoded.namespace_map import build_map, render_map
@@ -16,52 +20,78 @@ from uncoded.sync import sync_file
 DEFAULT_MAP_OUTPUT = Path(".uncoded/namespace.yaml")
 
 
-def _sync(*, check: bool = False) -> int:
+def _sync(*, root: Path | None = None, check: bool = False) -> int:
     """Sync (or verify) the namespace map, stub files, and instruction-file sections.
+
+    ``root`` is the directory the upward walk for ``pyproject.toml``
+    begins from; the parent of the located ``pyproject.toml`` becomes
+    the project anchor used to resolve source-root paths and to render
+    project-relative rel-paths in the namespace map and stubs.
+    Defaults to the current working directory at the CLI boundary.
+    Threading ``root`` through (rather than implicitly fetching cwd at
+    each downstream call site) keeps every working-directory dependency
+    visible in the signature and makes ``_sync`` testable against
+    arbitrary trees.
 
     When ``check=True``, the on-disk tree is not mutated: each step reports
     whether it would write. Returns 1 if any step reports a prospective
     change (so CI can gate on a stale index), 0 if the tree is already in
     sync. In apply mode, returns 0 on success or 1 on configuration error.
     """
-    cwd = Path.cwd()
+    if root is None:
+        root = Path.cwd()
+
+    pyproject_path = find_pyproject_toml(root)
+    if pyproject_path is None:
+        print(
+            "Error: No pyproject.toml found. "
+            "Add [tool.uncoded] source-roots to configure.",
+            file=sys.stderr,
+        )
+        return 1
+    project_root = pyproject_path.parent
+
     try:
-        source_roots = [r.resolve() for r in read_source_roots(cwd)]
-    except (FileNotFoundError, KeyError) as e:
+        source_roots = [
+            (project_root / r).resolve() for r in read_source_roots(project_root)
+        ]
+    except KeyError as e:
         print(f"Error: {e}", file=sys.stderr)
         return 1
 
-    for root in source_roots:
-        if not root.is_dir():
-            print(f"Error: {root} is not a directory", file=sys.stderr)
+    for src_root in source_roots:
+        if not src_root.is_dir():
+            print(f"Error: {src_root} is not a directory", file=sys.stderr)
             return 1
 
     changes = 0
 
-    modules = [m for root in source_roots for m in walk_source(root)]
+    modules = [
+        m for src_root in source_roots for m in walk_source(src_root, base=project_root)
+    ]
     map_content = render_map(build_map(modules))
     if sync_file(DEFAULT_MAP_OUTPUT, map_content, check=check):
         changes += 1
 
-    for root in source_roots:
-        changes += build_stubs(root, check=check)
+    for src_root in source_roots:
+        changes += build_stubs(src_root, base=project_root, check=check)
 
     # Dedupe configured instruction paths by resolved (canonical) path.
     # Without this, if CLAUDE.md is a symlink to AGENTS.md, pass 1 writes
     # through the symlink and reports the alias name while pass 2 finds
     # the file already in sync and reports nothing — asymmetric output
     # that hides the actual write target. Resolving collapses both aliases
-    # to the same canonical path, which we render relative to cwd for the
-    # user-facing line, falling back to the absolute resolved path when
-    # the file lives outside cwd.
+    # to the same canonical path, which we render relative to ``root`` for
+    # the user-facing line, falling back to the absolute resolved path when
+    # the file lives outside ``root``.
     seen_resolved: set[Path] = set()
-    for path in read_instruction_files(cwd):
+    for path in read_instruction_files(root):
         resolved = path.resolve()
         if resolved in seen_resolved:
             continue
         seen_resolved.add(resolved)
         try:
-            canonical = resolved.relative_to(cwd)
+            canonical = resolved.relative_to(root)
         except ValueError:
             canonical = resolved
         if sync_instruction_file(canonical, check=check):
--- a/src/uncoded/cli.py
+++ b/src/uncoded/cli.py
@@ -9,12 +9,12 @@ from uncoded.config import (
     read_instruction_files,
     read_source_roots,
 )
-from uncoded.extract import iter_source_files, walk_source
+from uncoded.extract import extract_modules, iter_source_files
 from uncoded.instruction_files import sync_instruction_file
 from uncoded.namespace_map import build_map, render_map
 from uncoded.serena_setup import setup
 from uncoded.skill import sync_skill
-from uncoded.stubs import build_stubs
+from uncoded.stubs import DEFAULT_STUBS_OUTPUT, _generate_stubs, _write_stubs
 from uncoded.sync import sync_file
 
 DEFAULT_MAP_OUTPUT = Path(".uncoded/namespace.yaml")
@@ -67,27 +67,28 @@ def _sync(*, root: Path | None = None, check: bool = False) -> int:
     changes = 0
 
     # Drive a single ``iter_source_files`` pass per source root and feed
-    # both the namespace map and the stubs pipeline from one read. Without
-    # this, each pipeline calls ``iter_source_files`` separately and a
-    # syntax-erroring file produces two identical ``warning: skipping``
-    # lines per ``uncoded sync`` invocation — the contract is one warning
-    # per broken file per sync.
+    # both the namespace map and the stubs pipeline from that one read.
+    # Without this, each pipeline runs ``iter_source_files`` separately
+    # and a syntax-erroring file produces two identical ``warning:
+    # skipping`` lines per ``uncoded sync`` invocation — the contract
+    # is one warning per broken file per sync.
     roots_with_files = [
         (src_root, list(iter_source_files(src_root, base=project_root)))
         for src_root in source_roots
     ]
 
     modules = [
-        m
-        for src_root, files in roots_with_files
-        for m in walk_source(src_root, base=project_root, files=files)
+        m for _src_root, files in roots_with_files for m in extract_modules(files)
     ]
     map_content = render_map(build_map(modules))
     if sync_file(DEFAULT_MAP_OUTPUT, map_content, check=check):
         changes += 1
 
     for src_root, files in roots_with_files:
-        changes += build_stubs(src_root, base=project_root, files=files, check=check)
+        stubs = _generate_stubs(files)
+        changes += _write_stubs(
+            stubs, src_root, DEFAULT_STUBS_OUTPUT, project_root, check=check
+        )
 
     # Dedupe configured instruction paths by resolved (canonical) path.
     # Without this, if CLAUDE.md is a symlink to AGENTS.md, pass 1 writes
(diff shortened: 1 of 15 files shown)


alimanfoo:
**Recommendation:** non-blocking concerns below, mostly around the working-directory contract not extending to writes.

## Non-blocking

1. **Reads anchor on `project_root`, but writes still resolve against cwd.** The PR description says "running `uncoded sync` from a subdirectory finds and writes the right targets," but the writes do not actually anchor: `sync_file(DEFAULT_MAP_OUTPUT, ...)` at `src/uncoded/cli.py:84`, `_write_stubs(..., DEFAULT_STUBS_OUTPUT, ...)` at line 89, `sync_instruction_file(canonical, ...)` at line 112 (where `canonical` is just `resolved.relative_to(project_root)`, i.e. a relative path that re-binds to cwd at write time), and `sync_skill(...)` at line 115 all write through cwd-relative paths. So an invocation from a subdirectory now reads source from `<project_root>/src/...` but writes `<subdir>/.uncoded/namespace.yaml`, `<subdir>/.uncoded/stubs/...`, `<subdir>/CLAUDE.md`, and `<subdir>/.claude/skills/...`. That is internally inconsistent — pre-PR the same invocation errored cleanly on the missing source root; post-PR it half-runs and splits the artifacts across two trees. Either anchor each write at `project_root` (`project_root / DEFAULT_MAP_OUTPUT`, etc., plus a `root=` for `sync_skill`), or keep the contract narrower than the description claims and adjust the PR body and `_sync` docstring accordingly.

2. **The new `root=` parameter on `_sync` is dormant.** The docstring says threading `root` through "makes `_sync` testable against arbitrary trees," but every test in `tests/test_cli.py` still relies on `_init_repo`'s `monkeypatch.chdir(tmp_path)` to set up cwd, and no test calls `cli._sync(root=...)` with anything other than the default. This is the same chip pattern issue #77 cited as the precedent (`#54`, `#60`: `setup(root=...)` added but only exercised via `chdir`). Until at least one test runs `_sync` from a cwd that is not the project root, the seam closing `#77` cites cannot actually be exercised end-to-end — which is also what would have surfaced finding 1 above.

3. **`extract_modules` and `_write_stubs` have no dedicated tests.** Both new functions are covered only indirectly through the legacy wrappers (`walk_source` for `extract_modules`, `build_stubs` for `_write_stubs`), and those wrappers now have no production callers — only tests. If the wrappers are kept just to preserve the test surface, that is a coverage gap rather than coverage; if they are dropped, the new pieces need their own tests.

4. **`test_skip_warning_emitted_once_per_broken_file` only checks stderr.** It verifies the warning count and prefix but does not assert that `good.py`'s stub was still written, which is the other half of the contract being defended (one warning, but pipeline still completes for non-broken files). A one-line `assert (tmp_path / ".uncoded" / "stubs" / "src" / "good.pyi").exists()` would make the test fail closed if a regression skipped both files.

## Nits

1. `cli.py` imports `_generate_stubs` and `_write_stubs` across modules despite the leading underscore — every other underscored symbol in the codebase is module-private. Either drop the prefix on the two callees (they are now part of the cli's actual surface area) or wrap them in a non-underscored public helper.

2. `_sync`'s docstring says `root` "defaults to the current working directory at the CLI boundary." It actually defaults to `None`, with the cwd substitution happening inside the body. Minor wording.

3. `_first_sentence`'s behaviour on whitespace-only docstrings has changed subtly — old code returned `""`, new returns `None`. Both are falsy, so downstream code is unaffected, but a one-line test would lock the new contract in.

## Out of scope but noticed

1. `walk_source` and `build_stubs` are now thin wrappers over the new pure functions and have no production callers — only test callers. Worth considering whether they earn their existence or whether the test suite should drop down to the new pieces directly so the wrappers can be deleted.

src/uncoded/cli.py
```diff
@@ -4,64 +4,111 @@
...
 
     changes = 0
 
-    modules = [m for root in source_roots for m in walk_source(root)]
+    # Drive a single ``iter_source_files`` pass per source root and feed
+    # both the namespace map and the stubs pipeline from that one read.
+    # Without this, each pipeline runs ``iter_source_files`` separately
+    # and a syntax-erroring file produces two identical ``warning:
+    # skipping`` lines per ``uncoded sync`` invocation — the contract
+    # is one warning per broken file per sync.
```
alimanfoo: This reads like information that was useful during the development session, but not useful to a later source code reader.

tests/test_cli.py
```diff
@@ -123,6 +123,73 @@ def test_error_when_source_root_missing(self, tmp_path, monkeypatch, capsys):
...
         assert "Error" in capsys.readouterr().err
 
+    def test_skip_warning_emitted_once_per_broken_file(
+        self, tmp_path, monkeypatch, capsys
+    ):
+        # A syntax-erroring source file must produce exactly one stderr
+        # warning per ``uncoded sync`` invocation. Before the dedup fix,
+        # ``walk_source`` and ``_generate_stubs`` each ran
+        # ``iter_source_files`` separately and each warned, so a single
+        # broken file produced two identical lines on stderr.
```
alimanfoo: This reads like comments relevant to the current development session but not useful for a later reader, who will not have any of the history and doesn't need to know it.

tests/test_cli.py
```diff
@@ -123,6 +123,73 @@ def test_error_when_source_root_missing(self, tmp_path, monkeypatch, capsys):
...
+
+    def test_root_param_anchors_reads_at_project_root_when_cwd_is_subdir(
+        self, tmp_path, monkeypatch
+    ):
+        # When ``_sync(root=...)`` is called with cwd inside a project
+        # subdirectory, the *read* anchor (source-root resolution and
+        # rel-path rendering in the namespace map) follows the project
+        # root derived from ``root``, not cwd. Output paths are still
+        # cwd-relative — that's documented as a separate concern, so
+        # this test only pins the read seam.
```
alimanfoo: Is this comment useful for a future code reader?

tests/test_cli.py
```diff
@@ -123,6 +123,73 @@ def test_error_when_source_root_missing(self, tmp_path, monkeypatch, capsys):
...
+
+        # Output paths are cwd-relative; namespace.yaml landed under the
+        # subdirectory we ran from. That's the documented behaviour.
+        namespace_path = tmp_path / "src" / ".uncoded" / "namespace.yaml"
+        assert namespace_path.exists()
+
+        # The rel-path inside is project-anchored: ``src/foo.py`` (from
+        # tmp_path), not ``foo.py`` (from cwd). If the read anchor had
+        # leaked back to cwd, the file would appear as a top-level
+        # ``foo.py`` key with no ``src/`` directory wrapper.
```
alimanfoo: Again, looks like historical info.

tests/test_stubs.py
```diff
@@ -155,6 +156,20 @@ def silent():
         module = extract_stub(source, "pkg/silent.py")
         assert module.functions[0].docstring_excerpt is None
 
+    def test_whitespace_only_docstring_yields_none(self):
+        # A whitespace-only docstring is observably a docstring at the
+        # AST level but carries no content. ``ast.get_docstring`` with
+        # the default ``clean=True`` cleans it to an empty string, which
+        # ``_first_sentence`` short-circuits as ``None`` rather than
+        # returning an empty excerpt.
```
alimanfoo: How useful is this comment?

tests/test_stubs.py
```diff
@@ -672,30 +685,78 @@ def _setup(self, tmp_path):
...
+        assert changes == 1
+        assert not (out / "src" / "foo.pyi").exists()
+
+    def test_prunes_orphan_stubs(self, tmp_path):
+        # Pre-existing stub under a subpackage of the source-root
+        # subtree whose source has been removed (modelled here by an
+        # empty stubs dict) should be deleted, and child directories
+        # left empty by the deletion should be pruned. The
+        # source-root stub directory itself is preserved — that's
+        # ``_write_stubs``'s "keep stubs_root itself" contract.
```
alimanfoo: How useful?

alimanfoo (commented):
General pattern here seems to be that there are a number of long and cryptic comments here, which look like information that was relevant during this particular coding session but will not be particularly useful for a future code reader, being mostly historical in nature.

PR #87, opened 2026-05-02
[claude] sync writes, _sync error UX, _first_sentence contract

### LINKED ISSUE #84: [claude] uncoded sync from a subdirectory writes artefacts to the subdirectory, not the project root
Running \`uncoded sync\` from a subdirectory of the project lands artefacts under the subdirectory rather than at the project root, even though source resolution correctly anchors at the project root since #77.

The asymmetry: \`cli._sync\` threads \`project_root\` through every project-relative read (source-root resolution, instruction-file lookup, rel-paths in the namespace map and stubs), but its writes still resolve relative to \`Path.cwd()\`:

- \`sync_file(DEFAULT_MAP_OUTPUT, ...)\` writes the namespace map at \`.uncoded/namespace.yaml\`.
- \`_write_stubs(..., DEFAULT_STUBS_OUTPUT, ...)\` writes stubs under \`.uncoded/stubs/\`.
- \`sync_instruction_file(canonical, ...)\` writes through \`canonical = resolved.relative_to(project_root)\`, a relative path that re-binds to cwd at write time.
- \`sync_skill(...)\` resolves its skill output paths relative to cwd.

A user running from \`<project>/src/\` reads source from \`<project>/src/...\` but writes \`<project>/src/.uncoded/namespace.yaml\`, \`<project>/src/.uncoded/stubs/\`, \`<project>/src/CLAUDE.md\`, etc. Pre-#77 the same invocation errored cleanly on a missing source root; post-#77 it half-runs, splitting artefacts across two trees.

Possible direction: anchor each write at \`project_root\` (\`project_root / DEFAULT_MAP_OUTPUT\` and similar), and consider a \`root=\` parameter on \`sync_skill\` analogous to the one \`_sync\` gained in #77. The \`_sync\` docstring already names the asymmetry as deferred work; this issue is the resolved tracking item.

### LINKED ISSUE #85: [claude] cli._sync's configuration-error messages quote KeyError literally and lack recovery hints
Two error messages in \`cli._sync\` (\`src/uncoded/cli.py\`) have small but visible UX problems that an end user encounters on first contact with a misconfigured \`pyproject.toml\`:

**1. KeyError quoting artifact.** When \`read_source_roots\` raises \`KeyError(\"No [tool.uncoded] source-roots found in pyproject.toml.\")\` and \`_sync\` does \`print(f\"Error: {e}\", file=sys.stderr)\`, the \`f\"{e}\"\` formatting calls \`KeyError.__str__\`, which returns the message wrapped in single quotes. The user sees:

\`\`\`
Error: 'No [tool.uncoded] source-roots found in pyproject.toml.'
\`\`\`

The literal quotes look like a formatting bug (which they are).

**2. Asymmetric recovery guidance.** The two adjacent configuration-error paths in \`_sync\` give different help levels:

- \`No pyproject.toml found.\` → followed by \`Add [tool.uncoded] source-roots to configure.\` (recovery hint provided)
- \`No [tool.uncoded] source-roots found in pyproject.toml.\` → no recovery hint, even though this is the more common case (the user has a \`pyproject.toml\` but hasn't added the uncoded section)

A user hitting the second message has no breadcrumb for what to do next.

Possible direction: format the message string directly in \`_sync\` (or use \`str(e)\` and have the readers raise plain strings instead of KeyError-wrapped strings) to drop the quoting; and add the same kind of recovery hint to the missing-section message that the missing-toml message already has.

### LINKED ISSUE #86: [claude] _first_sentence: third chip on the docstring-excerpt heuristic suggests an unresolved contract
The \`_first_sentence\` heuristic in \`src/uncoded/stubs.py\` has accumulated three chips in successive sessions, each patching a different symptom on the same surface:

- **#72 (closed):** \`_first_sentence\` truncated docstrings starting with \`e.g.\`, \`i.e.\`, \`Mr.\`, etc. Resolved by adding the \`\\s+[A-Z]\` lookahead so the period isn't treated as a sentence boundary unless followed by whitespace and a capital letter.
- **#81 (closed by #83):** the line-based fallback became narrowly reachable after the \`\" Z\"\` sentinel landed in #76, and the comment didn't say what it was for. Resolved by collapsing the function to a single regex with two alternatives: \`(.+?\\.(?=\\s+[A-Z])|.+?(?=\\n))\` against \`docstring.strip() + \"\\n\"\`.
- **Now:** the function's alt-2 path (\`.+?(?=\\n)\` — the newline-boundary fallback for period-less or no-capital-boundary docstrings) has no targeted test, and the companion \`if match is None: return None\` after \`re.match\` is unreachable in practice given the early-return on empty docstring.

Each chip patches a symptom on the same surface. The contract question worth asking explicitly:

- **What does \`_first_sentence\` promise its caller for input that has no \`.\` followed by \`\\s+[A-Z]\`?** The current implementation accepts it via alt-2 (returns the text up to the first newline), but no test pins this behaviour and the public docstring (\`\"Return the first sentence of a node's docstring, or None.\"\`) doesn't address the no-period case.
- **Which inputs reach which alt?** After #83's simplification, alt-1 covers period-bounded multi-sentence text and the abbreviation cases (\`e.g.\`, \`i.e.\`, \`U.S.\`); alt-2 covers single-sentence-with-period and period-less / no-capital-boundary inputs. The split is in the regex but not named at the docstring or test layer.

Naming the contract — at the docstring or the test layer — would let future contributors stop chipping at it.

Possible direction: pin the alt-2 contract with at least one test (e.g. a no-period docstring like \`\"\"\"Hello world\"\"\"\` returning \`\"Hello world\"\`), and either widen the function's docstring to name \"first sentence or first line, whichever comes first\" as the contract, or name \"first sentence\" explicitly and have alt-2 return \`None\` instead of the line, accepting information loss for no-terminator inputs.

alimanfoo:
This branch resolves three concerns surfaced in recent issues, each contained to its own commit.

**Sync writes from a subdirectory now land at the project root (#84).** Since #77 the read side of `uncoded sync` resolves project-relative paths against the discovered project root regardless of the user's working directory; the write side did not. Running from a subdirectory therefore split artefacts across two trees — source read from the project, but namespace map, stubs, instruction file, and skill outputs written under the subdirectory. A `root=` keyword is now threaded through every writer (`sync_file`, `remove_file`, `sync_instruction_file`, `sync_skill`, `_write_stubs`, `build_stubs`) so I/O resolves at the project root while printed paths stay project-relative for log readability. A behavioural test exercises the end-to-end: identical artefacts in identical locations whether sync runs from the project root or a subdirectory.

**Configuration-error messages no longer leak KeyError quoting and offer recovery guidance for both shapes (#85).** The "missing `[tool.uncoded]` section" path used to surface as `Error: 'No [tool.uncoded] source-roots found in pyproject.toml.'` — the literal single quotes were an artefact of `KeyError.__str__`. The same path also lacked the recovery hint that the adjacent "missing `pyproject.toml`" path provides, despite being the more common configuration error. `read_source_roots` now raises `LookupError` with the recovery hint baked into the message; both error shapes render symmetrically and without quoting noise.

**`_first_sentence` has a documented contract (#86).** The helper accumulated a chip in each of the last three sessions (#72, #81/#83, the current one), and the latest issue framed the recurrence as an unresolved contract — what does the function promise its caller for input that has no `.\s+[A-Z]`? The contract is now named in the function's docstring as "first sentence, or first line, whichever comes first," with the sentence-boundary heuristic and the abbreviation-truncation limit (`Mr. Smith arrived.`, `Dr. Jones`) made explicit. Tests pin both halves of the contract and the documented non-contract.

A small follow-up commit also collapses an unreachable `try/except` introduced by the GH84 anchoring work — the orphan-pruning display path's `relative_to` fallback was textually dead because `abs_stubs_root.rglob(...)` cannot yield paths outside `root` under `Path`'s join semantics.

Closes #84
Closes #85
Closes #86


alimanfoo:
**Recommendation:** approve subject to a couple of nits — no blocking concerns.

## Non-blocking

1. **`_first_sentence` invariant relies on `assert`** (`src/uncoded/stubs.py:108`). The old code had a defensive `if match is None: return None`; the new code asserts the invariant and immediately calls `match.group(1)`. The invariant is sound — `docstring.strip() + "\n"` always has at least one non-newline char before the trailing `\n`, so alt-2 (`.+?(?=\n)`) always matches — and the comment explains why. Worth knowing that under `python -O` the assert is stripped, so if a future edit ever broke the contract (e.g. removed the trailing `+ "\n"`, or weakened the early-return guard) the failure on optimized builds would be a bare `AttributeError` from `.group(1)` rather than the previous graceful `None` fallback. Behaviour-correct today; the trade-off is sharper failure on invariant break versus the old soft-fallback shape.

2. **Docstring readability nit on `_first_sentence`.** The sentence "Returns ``None`` only when the node has no docstring, or the docstring is whitespace-only (after `ast.get_docstring`'s `clean=True` normalisation reduces it to `""`)" packs the cause inside a parenthetical in the contract sentence. Splitting the parenthetical into a follow-on sentence reads more easily — but this is purely stylistic.

## Nits

1. **Redundant `root=` in `test_artefacts_match_when_run_from_subdir_vs_project_root`** (`tests/test_cli.py:240`). The second invocation passes `root=from_subdir` explicitly, short-circuiting the upward walk that real CLI users rely on. Since `_sync()` invoked from `from_subdir/src` would walk up and discover the same `project_root`, dropping the explicit `root=` would exercise the user-realistic "no flag from a subdir" path end-to-end. The current test still proves equivalence at the `_sync` API level, so this is a small thing.

2. **`Crucially,` adds nothing** (`tests/test_sync.py:100`). "Crucially, no write under cwd." reads cleanly as just "no write under cwd."

## Out of scope but noticed

1. **`cli._sync`'s "is not a directory" error path leaks an absolute path and lacks a recovery hint** (`src/uncoded/cli.py:65`). `print(f"Error: {src_root} is not a directory", ...)` formats `src_root = (project_root / r).resolve()`, which is absolute, while GH85 just polished two sibling configuration-error messages to be project-relative-friendly with recovery hints. This is the third configuration-error path in `_sync` and the only one without that polish. Pre-existing; flagging because GH85's work makes it visible.

2. **The `root: Path | None = None` parameter is duplicated across five writers** with the same `target = root / path if root is not None else path` line and a near-identical docstring paragraph each (`sync_file`, `remove_file`, `sync_instruction_file`, `sync_skill`, `_write_stubs`/`build_stubs`). A small private helper plus one canonical contract paragraph that the docstrings reference would shrink the duplication. Surface-only — no behavioural concern — but the per-callsite repetition is now visible in five places.

3. **Assertion-as-documentation pattern.** Adjacent to finding 1 above, but at the codebase level: this PR introduces an `assert` whose body is invariant-checking, immediately followed by use of the value the assertion vouches for. If this pattern is intended as a project-wide convention for internal invariants, a one-line note in `CLAUDE.md` (or the closest stylistic equivalent) would help future contributors recognise it as a deliberate idiom rather than dead defensive code that can be deleted.

PR #91, opened 2026-05-04
[claude] privatise build_stubs and tighten _write_stubs signature

### LINKED ISSUE #89: [claude] build_stubs is shaped like public API but only tests call it
`build_stubs` in `src/uncoded/stubs.py` is shaped like a public convenience wrapper — no underscore prefix, full docstring, `Optional` parameter defaults — but it's unused outside of `tests/test_stubs.py`. `cli._sync` reaches into the private `_write_stubs` directly so it can share `roots_with_files` across the namespace-map and stubs steps.

The `#84` work added a `root=` parameter to `build_stubs` for signature symmetry with `_write_stubs`, expanding the public surface of a function that has no in-package consumer.

Prior issue on this shape: #54 (`setup(root=...)` was identified as a dormant public API; the dormant parameter was eventually addressed).

- Is `build_stubs` real public API (consumed by users importing `uncoded` as a library), or has it grown into a convenience wrapper that only tests exercise?
- If public: should the cli use it instead of reaching into `_write_stubs`? That would make production exercise the same code path as tests, eliminating the test-vs-production drift.
- If test-only: should it be private (or moved to a test-utility module), and should the per-callsite parameter surface (now including `root=`) be trimmed accordingly?

Possible direction: confirm whether the function has any external consumers, then either (a) refactor `_sync` to use `build_stubs` so production exercises the public API, or (b) make it private (`_build_stubs`) and trim the parameter surface.

alimanfoo:
## Why

`build_stubs` was shaped like public API — no underscore, full docstring, optional defaults — but had no caller outside the test suite. The shape lied; the function is an internal end-to-end helper used by tests. This PR renames it to `_build_stubs` and tightens its signature to match its actual usage.

While we were in `stubs.py`, we applied the same private-function discipline (keyword-only signature, no defaults at the definition, every argument stated at the call site) to `_write_stubs`, the only other private function in the module.

Closes #89.

### DIFF
--- a/.uncoded/namespace.yaml
+++ b/.uncoded/namespace.yaml
@@ -113,7 +113,7 @@ src/:
       render_stub:
       _generate_stubs:
       _write_stubs:
-      build_stubs:
+      _build_stubs:
     sync.py:
       sync_file:
       remove_file:
--- a/.uncoded/stubs/src/uncoded/stubs.pyi
+++ b/.uncoded/stubs/src/uncoded/stubs.pyi
@@ -69,7 +69,7 @@ def _generate_stubs(files: Iterable[tuple[str, str]]) -> dict[Path, str]:
     """Return a mapping from stub relative paths to rendered stub content."""
     ...
 
-def _write_stubs(stubs: dict[Path, str], source_root: Path, output_dir: Path, base: Path, *, root: Path | None, check: bool) -> int:
+def _write_stubs(*, stubs: dict[Path, str], source_root: Path, output_dir: Path, base: Path, root: Path | None, check: bool) -> int:
     """Write *stubs* under *output_dir* and prune orphans under *source_root*."""
     ...
 
--- a/.uncoded/stubs/src/uncoded/stubs.pyi
+++ b/.uncoded/stubs/src/uncoded/stubs.pyi
@@ -73,7 +73,7 @@ def _write_stubs(stubs: dict[Path, str], source_root: Path, output_dir: Path, ba
     """Write *stubs* under *output_dir* and prune orphans under *source_root*."""
     ...
 
-def build_stubs(source_root: Path, output_dir: Path, base: Path | None, *, root: Path | None, check: bool) -> int:
+def _build_stubs(*, source_root: Path, output_dir: Path, base: Path, check: bool) -> int:
     """Sync stub files for all symbols under source_root, removing any orphans."""
     ...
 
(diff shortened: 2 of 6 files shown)


alimanfoo:
## Recommendation: approve

PR #91 cleanly addresses #89. `build_stubs` had a public-API shape (no underscore, optional defaults, broad signature) with no consumer outside the test suite; this change privatises it (`_build_stubs`), drops the unused `root=` plumbing, and tightens both `_build_stubs` and `_write_stubs` to keyword-only signatures with no defaults. Every call site is updated (CLI, tests, regenerated stubs and namespace map). Full test suite passes locally (204/204) and `uv run uncoded check` is clean.

The diff is contained, the rename is mechanical, and the function shapes now match their actual use: `_write_stubs` takes a required `root: Path | None` (the CLI passes `project_root`; the test helper passes `None`), and `_build_stubs` drops `root=` entirely since no caller anchored it elsewhere.

## Findings

### Blocking

None.

### Non-blocking

None.

### Nits

1. **Test class names still say `TestBuildStubs` / `TestBuildStubsCheckMode`** (`tests/test_stubs.py:622`, `:718`). The class docstrings inside got updated to reference `_build_stubs`, but the class names did not. Not worth a rename — class-name churn has low value here, and the canonical name is carried by the docstring and the imports. Mentioning only because the rest of the diff was thorough about the rename.

## Out of scope but noticed

1. **`_write_stubs` `root` parameter is now exercised in two opposite directions by the two callers.** `cli._sync` always passes `project_root` (non-`None`); `_build_stubs` always passes `None`. The branch in `_write_stubs` (`anchored = root / stub_path if root is not None else stub_path`, `display = existing.relative_to(root) if root is not None else existing`) is therefore real and used, but each caller exercises only one half. Worth noting because GH54 tracked a similar shape on `setup(root=...)` where the parameter was dormant; here it's not dormant, just asymmetric. The asymmetry is honest about the difference between production (CLI, anchored) and test-helper (cwd-relative) use, but if production ever wanted the cwd-relative branch, only tests would have exercised it. No action needed in this PR.

2. **`.uncoded/reviews/2026-04-25-001215.md` still references the old `build_stubs` name** (lines 24, 31, 34, 38, 42, 46, 49) with old line numbers and the pre-#84 signature. These look like archived past-review artefacts rather than living documentation, so the staleness is expected — flagging in case the convention is to update them on rename.

PR #93, opened 2026-05-04
[claude] drop "in source order" claim from index header and agent prose

### LINKED ISSUE #92: [claude] namespace.yaml header claims source order but module constants are hoisted
The header generated into `.uncoded/namespace.yaml` claims source order:

> \# Entries within a file or class appear in source order.

But the renderer groups module-level constants ahead of classes and functions, regardless of their position in source. Visible in `src/uncoded/stubs.py`:

- Source: `VALUE_WIDTH_CAP` at line 14, then classes (`StubParam` etc.) starting around line 25, then helpers, then `DEFAULT_STUBS_OUTPUT` at line 372 (between `_generate_stubs` and `_write_stubs`).
- Namespace map for `stubs.py`: both constants listed first, classes next, functions last.
- Generated stub at `.uncoded/stubs/src/uncoded/stubs.pyi`: same — `VALUE_WIDTH_CAP` and `DEFAULT_STUBS_OUTPUT` grouped at lines 13–14, then `_first_sentence` at line 16.

So the index's two artefacts (namespace map and stubs) agree with each other and disagree with the source, while the header documents a contract that doesn't match either renderer's behaviour.

Surfaced in the post-merge sweep for PR #91. Pre-existing — `DEFAULT_STUBS_OUTPUT`'s mid-module placement traces back to commit 550d3f56, before the session.

- Is "constants first, then classes, then functions in source order" the intended layout for the index? If so, the header's "source order" claim is loose and should describe what actually happens.
- Or is true source order the intended layout, and the renderer is wrong to hoist?
- Either way, source `stubs.py` is the only file in this repo where a module-level constant lives mid-module — moving `DEFAULT_STUBS_OUTPUT` next to `VALUE_WIDTH_CAP` would dissolve the per-file divergence regardless of how the index contract is settled.

Possible direction: pick one interpretation and align the other two artefacts. Either move `DEFAULT_STUBS_OUTPUT` to the top of `src/uncoded/stubs.py` and keep the header as-is, or update the header (and possibly the renderer's docstring) to describe the hoist explicitly.

alimanfoo:
Closes #92.

## Summary

The header in `.uncoded/namespace.yaml` claimed that "Entries within a file or class appear in source order," but the renderer hoists module-level constants ahead of classes and functions regardless of their position in source. Issue #92 surfaced this divergence.

After scoping discussion, the fix is **not** to align the renderer to the header, and **not** to move source code in this repo (uncoded is a tool used in arbitrary downstream codebases — we cannot tune source layout to match the renderer). Instead: the order claim has no consumer. The namespace map is loaded into agent context all at once; agents do not read it sequentially. Order is decorative. Drop the claim.

## Changes

Three template edits, four regenerated artefacts:

- `src/uncoded/namespace_map.py` — drop the two-line "Entries within a file or class appear in source order." sentence from the YAML header.
- `src/uncoded/instruction_files.py` — remove the "— in source order" qualifier from the agent-facing prose; the surrounding "Read it once, in full, at session start." wording is preserved.
- `src/uncoded/skill.py` — remove the "— in source order" qualifier from the coherence-review skill template; "Read … in full" and "Do not skim" are preserved.
- `.uncoded/namespace.yaml`, `AGENTS.md`, `.claude/skills/coherence-review/SKILL.md`, `.agents/skills/coherence-review/SKILL.md` — regenerated by `uv run uncoded sync` to match.

The renderer's actual ordering behaviour is untouched. The contract being preserved (and reinforced): agents are told to read the namespace map *in full*.

## Test plan

- [x] `uv run pre-commit run --all-files` — all hooks pass, including the `uncoded` index check.
- [x] `uv run pytest` — 204 passed.
- [x] Post-edit `grep -rn "source order"` returns zero hits across the tree.
- [x] Header tests in `tests/test_namespace_map.py` (`test_header_appears_at_top`, `test_header_mentions_stub_pointer`, `test_header_does_not_break_yaml_parse`) lock the surfaces that still hold and remain valid.


alimanfoo:
## Review of #93

**Recommendation: approve.**

This is a tight, well-scoped documentation fix. The header in `.uncoded/namespace.yaml` claimed an ordering contract the renderer does not in fact provide, the framing in #92 correctly identified that the order claim has no consumer, and the diff drops the claim from all three template sources in one place each, then regenerates the four artefacts. CI is green, including the `uncoded` index check.

I read #92, the PR description, the diff, and the surrounding source for `namespace_map.py`, `instruction_files.py`, `skill.py`, and the regenerated artefacts. I also did a tree-wide search for any leftover "source order" / "in source order" wording — no hits.

### Findings

#### Blocking
None.

#### Non-blocking
None.

#### Nits
None.

### Out of scope but noticed

Two pre-existing items the diff drew my eye to but does not itself address. Flagging for post-merge triage; not part of this PR's scope.

1. **`tests/test_namespace_map.py:104` — `test_source_order_preserved` is misleadingly named.** The test passes a `ModuleInfo` with `classes=[Alpha]`, `functions=[zebra, apple]` and asserts the keys come out as `["Alpha", "zebra", "apple"]`. That is insertion-order preservation across the build (classes-then-functions, then within each list as supplied), not source-line order — and the very next test, `test_constants_precede_classes_and_functions`, locks in that constants are hoisted ahead of source position. With the "source order" claim now dropped from the header and the agent prose, the test name is the last public-facing place the loose phrase persists. A rename to something like `test_preserves_class_then_function_insertion_order` would close the loop, but that is contract-name polish on an internal test, not a behaviour gap.

2. **`src/uncoded/stubs.py` — `DEFAULT_STUBS_OUTPUT` still lives mid-module at line 372.** #92 noted this incidentally as the only file in the repo where a module-level constant is not at the top, and observed that the renderer hoists it in both the namespace map and the stub. The PR (correctly, per the scoping discussion) does not move the constant — uncoded is a tool used in arbitrary downstream codebases, so source layout is not ours to police. But within this repo, moving `DEFAULT_STUBS_OUTPUT` next to `VALUE_WIDTH_CAP` at line 14 would make the source layout match the rendered layout for this one file at zero cost. Worth a moment of triage on whether to tidy.

PR #96, opened 2026-05-05
[claude] cli._sync source-root error UX + read_source_roots contract (#88)

### LINKED ISSUE #88: [claude] cli._sync's configuration-error reporting splits across three shapes
After #85, three configuration-error sites in `cli._sync` use three different idioms for reaching the user, with three different polish levels:

1. `find_pyproject_toml` returns `None` when no `pyproject.toml` is found. `_sync` checks the sentinel and prints `Error: No pyproject.toml found.\nAdd [tool.uncoded] source-roots to configure.` — recovery hint included.
2. `read_source_roots` raises `LookupError` when `[tool.uncoded] source-roots` is missing. `_sync` catches and prints `f"Error: {e}"`. The message (with recovery hint) lives inside the exception.
3. `_sync` itself prints `f"Error: {src_root} is not a directory"` for the source-root directory check (`src/uncoded/cli.py:65`). `src_root = (project_root / r).resolve()` is absolute, so the message leaks an absolute path; there's no recovery hint.

There's also a duplication: the same `Add [tool.uncoded] source-roots to configure.` text exists at the cli call site (path 1) and inside `read_source_roots`'s `FileNotFoundError` raise. The `FileNotFoundError` raise in `read_source_roots` is unreachable from `_sync` (since `_sync` calls `find_pyproject_toml` itself first and short-circuits), so the duplication has no current consumer either.

Prior chip on this surface: #85.

- Where should configuration-error messages live — at the call site (`cli.py`), inside the readers (`config.py`), or in a shared error type?
- Is sentinel-return-then-check or raise-then-catch the canonical idiom for this codebase's config-error reporting?
- What's the format contract — does every config-error message render project-relative-friendly with a recovery hint, or only the ones we've polished?

Possible direction: pick one canonical idiom (raise-then-catch is more uniform now that #85 has moved one site there), polish the third site (`is not a directory`) to match, and remove the dead `FileNotFoundError` raise from `read_source_roots`.

alimanfoo:
Closes #88.

## Summary

GH88 surfaced three configuration-error sites in `cli._sync` with three different idioms. Rather than unify them under a single contract (which would add structure for symmetry, not for users), this PR resolves the tangle by:

- **Polishing the third site's UX** so it matches the other two: project-relative path (no absolute-path leak), recovery hint pointing at `[tool.uncoded] source-roots`.
- **Removing a duplicate `pyproject.toml` walk** between `cli._sync` and `read_source_roots`. `read_source_roots` now takes the located `pyproject_path` directly. This also dissolves a "dead" `FileNotFoundError` branch that was unreachable at runtime but load-bearing for the type checker — both go away in one move, no `# ty: ignore`/`cast`/`assert` needed.

The remaining signature asymmetry across the three readers (`start: Path` for `read_project_name` and `read_instruction_files`, `pyproject_path: Path` for `read_source_roots`) reflects a real contract distinction: the first two have a defined fallback when no `pyproject.toml` is found; the third does not.

## Changes

- `src/uncoded/cli.py`: third error message now uses the configured (project-relative) path and includes a recovery hint; passes `pyproject_path` to `read_source_roots`.
- `src/uncoded/config.py`: `read_source_roots(pyproject_path: Path) -> list[Path]` — opens the toml directly; raises `LookupError` only on missing `[tool.uncoded] source-roots`.
- `tests/test_cli.py`: tightens `test_error_when_source_root_missing` to assert problem statement, recovery hint, and that the absolute `tmp_path` doesn't appear in stderr.
- `tests/test_config.py`: `TestReadSourceRoots` updated to pass `pyproject_path`; `test_raises_if_no_pyproject_toml` removed (the contract no longer exists).
- `.uncoded/`: index regenerated to reflect the signature and test changes.

## Test plan

- [x] `uvx ruff check src tests` — passes
- [x] `uvx ruff format --check src tests` — passes
- [x] `uvx ty check src tests` — passes
- [x] `uv run pytest` — 203 passed (was 204; one fewer because the unreachable-branch test was removed)
- [x] `uv run uncoded check` — clean

### DIFF
--- a/.uncoded/namespace.yaml
+++ b/.uncoded/namespace.yaml
@@ -156,7 +156,6 @@ tests/:
       test_falls_back_to_start_name_when_name_missing:
     TestReadSourceRoots:
       test_reads_source_roots:
-      test_raises_if_no_pyproject_toml:
       test_raises_if_no_uncoded_section:
     TestReadInstructionFiles:
       test_returns_default_when_no_pyproject_toml:
--- a/.uncoded/stubs/src/uncoded/config.pyi
+++ b/.uncoded/stubs/src/uncoded/config.pyi
@@ -14,8 +14,8 @@ def read_project_name(start: Path) -> str:
     """Read the project name, falling back to the start-dir name."""
     ...
 
-def read_source_roots(start: Path) -> list[Path]:
-    """Read source roots from ``[tool.uncoded] source-roots`` in ``pyproject.toml``."""
+def read_source_roots(pyproject_path: Path) -> list[Path]:
+    """Read source roots from ``[tool.uncoded] source-roots``."""
     ...
 
 def read_instruction_files(start: Path) -> list[Path]:
--- a/.uncoded/stubs/tests/test_config.pyi
+++ b/.uncoded/stubs/tests/test_config.pyi
@@ -35,9 +35,6 @@ class TestReadSourceRoots:
     def test_reads_source_roots(self, tmp_path):
         ...
 
-    def test_raises_if_no_pyproject_toml(self, tmp_path):
-        ...
-
     def test_raises_if_no_uncoded_section(self, tmp_path):
         ...
 
--- a/src/uncoded/cli.py
+++ b/src/uncoded/cli.py
@@ -53,7 +53,7 @@ def _sync(*, root: Path | None = None, check: bool = False) -> int:
     project_root = pyproject_path.parent
 
     try:
-        configured_roots = read_source_roots(project_root)
+        configured_roots = read_source_roots(pyproject_path)
     except LookupError as e:
         print(f"Error: {e}", file=sys.stderr)
         return 1
--- a/src/uncoded/cli.py
+++ b/src/uncoded/cli.py
@@ -53,16 +53,19 @@ def _sync(*, root: Path | None = None, check: bool = False) -> int:
     project_root = pyproject_path.parent
 
     try:
-        source_roots = [
-            (project_root / r).resolve() for r in read_source_roots(project_root)
-        ]
+        configured_roots = read_source_roots(project_root)
     except LookupError as e:
         print(f"Error: {e}", file=sys.stderr)
         return 1
+    source_roots = [(project_root / r).resolve() for r in configured_roots]
 
-    for src_root in source_roots:
+    for configured, src_root in zip(configured_roots, source_roots, strict=True):
         if not src_root.is_dir():
-            print(f"Error: {src_root} is not a directory", file=sys.stderr)
+            print(
+                f"Error: source root {configured} is not a directory. "
+                "Check [tool.uncoded] source-roots in pyproject.toml.",
+                file=sys.stderr,
+            )
             return 1
 
     changes = 0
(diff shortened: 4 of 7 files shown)


alimanfoo:
**Recommendation:** Looks good — small, targeted, and the rejection of the "unify" framing is well argued in the description. A couple of non-blocking observations and a nit below.

## Non-blocking

1. **`read_source_roots` no longer documents what it requires of its caller.** `src/uncoded/config.py:45-51`. The new docstring says it "Reads the section from the given `pyproject.toml`" and raises `LookupError` when the section is missing — but it doesn't say the caller must pass a path to a file that exists and is readable. Inside `_sync`, that invariant is met by calling `find_pyproject_toml` first; from outside, a programmatic caller passing a non-existent path now gets a raw `FileNotFoundError` from `Path.open` rather than a tidy `LookupError`. A one-line note saying the caller must supply an existing readable file would make the contract explicit and match the *why* the PR description gives for the signature change.

2. **The recovery-hint wording diverges across the three sites in a small but visible way.** `src/uncoded/cli.py:48-49`, `src/uncoded/config.py:58-61`, `src/uncoded/cli.py:65-67`. Sites one and two say `Add [tool.uncoded] source-roots to configure.`; site three says `Check [tool.uncoded] source-roots in pyproject.toml.` "Add" vs "Check" is a deliberate distinction (you add a missing section, you check an existing one), so the divergence is defensible — but it is a small inconsistency a user may notice if they hit two of the three errors in sequence. Worth registering as a conscious call rather than fixing.

## Nits

1. **Test polish across the three error sites is now uneven.** `tests/test_cli.py:106-109` (`test_error_when_no_pyproject_toml`) still only asserts `"Error" in err`, while sites two (`test_error_when_uncoded_section_missing`) and three (`test_error_when_source_root_missing`, tightened in this PR) now assert problem statement, recovery hint, and absence of absolute-path leakage. The first test is the loosest of the three. Tightening it to match would close the gap; left as-is, the asymmetry mirrors the asymmetry the PR is otherwise polishing out.

2. **`zip(configured_roots, source_roots, strict=True)` defends an invariant the same function just enforced two lines earlier.** `src/uncoded/cli.py:60-62`. `source_roots` is built from `configured_roots` by a comprehension immediately above, so they are guaranteed equal-length by construction. `strict=True` is harmless but reads as defensive. Iterating the configured paths once and resolving inline in the loop body would be simpler; not worth changing if the explicit pairing reads more clearly to you.

alimanfoo:
Addressed review comments. Three follow-on commits on the branch:

- **c8a24fa** — `read_source_roots` docstring now states the caller invariant (must supply an existing readable file). Addresses non-blocking 1.
- **f5c3ae9** — `test_error_when_no_pyproject_toml` tightened to mirror the other two error-site tests (problem statement, recovery hint, no path leak). Addresses nit 1.
- **36ac147** — `_sync` source-root resolve+check folded into a single loop; `zip(..., strict=True)` retired (the 1:1 invariant it defended was structural after the change). Addresses nit 2.

Non-blocking 2 (Add vs Check recovery-hint wording at the three sites) is a deliberate distinction — `Add` for "section is missing", `Check` for "section is present but the path it lists is wrong." Conscious call rather than a fix; not changing it.

PR #98, opened 2026-05-07
[claude] one project anchor across the sync pipeline (#90)

### LINKED ISSUE #90: [claude] _sync's project-anchor vocabulary names several concepts at once
The `cli._sync` pipeline uses several variables named `root`, `project_root`, and `base` that are conceptually project-anchor-ish but operationally distinct:

- `_sync(*, root: Path | None = None, ...)` — `root` is the *upward-walk start* used to find `pyproject.toml`. Defaults to cwd at the CLI boundary.
- `project_root` (local in `_sync`) — the parent directory of the located `pyproject.toml`.
- `sync_file(*, root=...)`, `remove_file(*, root=...)`, `sync_instruction_file(*, root=...)`, `sync_skill(*, root=...)`, `_write_stubs(*, root=...)`, `build_stubs(*, root=...)` — `root` here is the *filesystem-I/O anchor* for project-relative paths.
- `_write_stubs(..., base, ...)` and `build_stubs(..., base=...)` — `base` is the *rel-path render anchor* used when stub-file headers print project-relative paths.

In production all four collapse to the same value (cwd ⇒ `project_root` ⇒ I/O anchor ⇒ rel-path base). At the only production call site, `_sync` passes `project_root` to `_write_stubs` twice — once positionally as `base`, once as `root=`.

Prior chips on this surface: #54, #60, #68, #77, #84 — five closed issues across the project-anchor / cwd / `setup`-vs-`_sync` family. Three in-session vantage points during the #84 work (per-task review, PR review, post-merge sweep) named the same vocabulary collapse without converging on a single fix.

- Is "project anchor" one concept with four operational forms, or four genuinely distinct concepts that happen to collapse in production?
- If one concept: are the four parameter names (`root` for walk-start, `root` for I/O, `base` for rel-paths, the local `project_root`) accidentally three names for one thing?
- Where in the pipeline does the anchor "settle" — is `project_root` the canonical form, and everything else either feeds into it or is derived from it?

Possible direction: small renames have been chipped at across five issues without resolving the underlying confusion. A structural seam may be needed — either rename writer `root=` to `project_root=` so the threading is named, introduce a small context object that carries the anchor explicitly, or accept the collapse and document `_sync` as the only place where the four forms diverge.

alimanfoo:
## Summary

The `_sync` pipeline used to thread the project anchor through four
parameter slots — `_sync(root=...)`, a local `project_root`, writer
`root=` arguments, and `_write_stubs(base=...)`. All four collapsed to
the same value in production, but the signatures pretended otherwise:
`_write_stubs` took both `base` and `root`, and `cli._sync` passed
`project_root` to it twice.

This PR names what's actually being threaded. Every writer
(`sync_file`, `remove_file`, `sync_instruction_file`, `sync_skill`,
`_write_stubs`) takes `project_root: Path | None` for the project
anchor. `_sync`'s own parameter — which names the directory the
upward walk for `pyproject.toml` begins from, not the anchor — is
renamed `start`. `_write_stubs`'s old `base` and `root` parameters
collapse into one non-optional `project_root: Path`.

Concretely:

- `_write_stubs(base=..., root=...)` → `_write_stubs(project_root=...)`,
  with `Path` (no longer `Path | None`); the `Path.cwd()` fallback
  branch was unreachable.
- `sync_file`, `remove_file`, `sync_instruction_file`, `sync_skill`:
  `root=` → `project_root=`, with all call sites and docstrings
  updated.
- `cli._sync(root=...)` → `cli._sync(start=...)`, with the docstring
  rewritten to drop the four-form vocabulary and name `start` and
  `project_root` plainly.
- The four `*RootAnchor` test classes and the `test_root_anchors_*` /
  `test_root_*` methods are renamed to match.

The behaviour is unchanged. CI (lint, format, ty, full test suite, and
the project's own `uncoded check`) is the freshness gate.

Closes #90.


alimanfoo:

The diff cleanly resolves the four-name vocabulary collapse: `_write_stubs` loses its duplicated `base`/`root` pair, every writer now takes `project_root`, and `_sync`'s own parameter is renamed `start` to mark it as the upward-walk seed rather than the anchor. The fact that `cli._sync` used to pass `project_root` twice into `_write_stubs` was the smoking gun, and the fix is exactly that. Stubs, namespace map, and tests are kept in lock-step. CI is the freshness gate.

## Non-blocking

1. `sync_instruction_file` and `sync_skill` keep `project_root: Path | None = None`, but in production they are only ever called from `cli._sync`, which always passes `project_root`. The same "fallback branch is unreachable" argument the PR uses to narrow `_write_stubs` to a non-optional `Path` could be applied to these two writers. `sync_file`/`remove_file` are general-purpose and have many test call sites without the kwarg, so their `Path | None` shape is justified; the two narrower writers are not. Worth a follow-up if you want the simplification to extend.

## Nits

1. `src/uncoded/stubs.py:390-391` — the `_write_stubs` docstring still hedges with "``output_dir`` is *treated as* relative to it for filesystem I/O." Now that `project_root` is non-optional, `output_dir` is *always* anchored at `project_root` — no second mode to hedge against. A flatter phrasing ("``output_dir`` is anchored at ``project_root`` for filesystem I/O...") would match the new contract more directly.

## Out of scope but noticed

1. `tests/test_cli.py:202` — the test asserts `cli._sync(start=tmp_path) == 0`, but the test name (`test_anchors_reads_and_writes_at_project_root_when_cwd_is_subdir`) and its body still describe the project-anchor behaviour, not the upward-walk start. The new vocabulary makes that drift visible: `start` is what the test passes; `project_root` is what the test verifies. Pre-existing; just noted for the post-merge sweep.
2. `src/uncoded/instruction_files.py:171-173` — the `sync_instruction_file` docstring describes the `project_root` anchor in terms of "filesystem I/O" vs "project-relative display." That's the same boilerplate as `sync_file`/`remove_file`/`sync_skill`. The four docstrings now share a near-identical paragraph; if the contract drifts again, four sites will need to drift together. A shared phrase or a cross-reference could prevent that, but it's a future-proofing concern only.

src/uncoded/stubs.py
```diff
@@ -377,19 +377,22 @@ def _write_stubs(
...
-    cleanup is skipped.
+    :func:`_generate_stubs`.
+
+    ``project_root`` is the single project anchor and must already be
+    resolved. ``output_dir`` is anchored at ``project_root`` for
+    filesystem I/O while printed messages remain project-relative, and
+    the orphan-cleanup subtree is anchored at
+    ``output_dir / source_root.relative_to(project_root)`` (so
+    ``project_root`` must be an ancestor of ``source_root`` for cleanup
+    to run; otherwise cleanup is skipped).
```
alimanfoo: This is utterly incomprehensible.

src/uncoded/sync.py
```diff
@@ -27,14 +27,15 @@ def sync_file(
...
-    the current working directory while keeping log lines project-relative.
-    Passing an absolute ``path`` makes ``root`` a no-op (the absolute side
-    wins under ``Path``'s join semantics).
+    When ``project_root`` is provided, ``path`` is treated as relative to
+    ``project_root`` for filesystem I/O while the printed message still
+    names ``path`` as given. This lets callers anchor writes at a project
+    root regardless of the current working directory while keeping log
+    lines project-relative. Passing an absolute ``path`` makes
+    ``project_root`` a no-op (the absolute side wins under ``Path``'s
+    join semantics).
```
alimanfoo: This is just about comprehensible but still pretty complex.

PR #101, opened 2026-05-07
[claude] one project-anchor name across reads, writes, and prose (#99)

### LINKED ISSUE #99: [claude] project-anchor surface still has scattered drift after #90's writer-side collapse
#90 collapsed the project-anchor vocabulary on the writer family: `_write_stubs`, `sync_file`, `remove_file`, `sync_instruction_file`, and `sync_skill` all converged on a single `project_root` parameter. But the project-anchor surface is wider than the writers, and three structurally-related residues survived the collapse — each small, but each on the same contract that prior chips have already named:

1. **`iter_source_files(source_root, base: Path | None = None)` in `src/uncoded/extract.py:101` is the read-side outlier.** It's the last public function on the sync pipeline still using `base` for what every writer now names `project_root`. `cli._sync` calls it as `iter_source_files(src_root, base=project_root)` at `src/uncoded/cli.py:76` — same value, different name, on the same call chain.

2. **Five docstrings name the same `project_root` parameter inconsistently.** `sync_file` and `remove_file` got a plain-English rewrite in #98 task 9. `sync_instruction_file` and `sync_skill` still carry the older "treated as relative to" wording from #98 task 2. `_build_stubs`'s docstring describes its `project_root` use yet again, in a third register. Five sites, three voices, one parameter.

3. **`DEFAULT_MAP_OUTPUT` and `DEFAULT_STUBS_OUTPUT` are constants-as-documentation.** Each is referenced exactly once at a single call site inside `_sync`. The `DEFAULT_` prefix promises an alternative form that doesn't exist — there is no `--map-output` or `--stubs-output` CLI flag. They thread through the writer pipeline against `project_root`, which is why they belong on this surface.

**Why one issue, not three.** Six prior chips on this family — #54, #60, #68, #77, #84, #90 — landed scattered fixes on scattered symptoms. #90 itself called out that \"small renames have been chipped at across five issues without resolving the underlying confusion.\" The project-anchor surface is the contract; the recurrence pattern is itself the behaviour gap. Filing three separate small issues lets the next three sessions each fix the easiest one and still leave the contract un-named. Filing one names it.

**Suggested direction.** A single converged answer to \"what is the project-anchor surface, what gets to ride on it, and what does each writer/reader/constant promise about it?\" — not three independent renames. The shape of the answer is open: it might be propagating `project_root` as the universal name (resolving residue 1), reducing the four near-identical writer paragraphs to a shared description (residue 2), and removing the unused `DEFAULT_` prefix or expressing those constants as inlined call-site values (residue 3) — all in one pass with one set of trade-offs.

alimanfoo:
## Summary

After #90 collapsed the project-anchor vocabulary on the writer family
(`sync_file`, `remove_file`, `sync_instruction_file`, `sync_skill`,
`_write_stubs` all converged on `project_root`), three residues
remained on the same surface:

- The read-side functions `iter_source_files` and `walk_source` still
  named the same anchor `base`. The CLI call site spelled it
  `iter_source_files(src_root, base=project_root)` — same value,
  different name, on the same call chain.
- Five docstrings described the same `project_root` parameter in at
  least three different voices.
- `DEFAULT_MAP_OUTPUT` and `DEFAULT_STUBS_OUTPUT` were referenced once
  each at single call sites in `_sync`, with a `DEFAULT_` prefix that
  promised a `--map-output` / `--stubs-output` flag that doesn't exist.

#99 framed these three as a recurrence — six prior chips on this
family across #54, #60, #68, #77, #84, #90 — and asked for a single
converged answer rather than three small fixes.

This PR delivers that answer in three commits:

- **read-side rename.** `base` → `project_root` on `iter_source_files`
  and `walk_source`, plus the one CLI call site and six test call
  sites. No logic changes.
- **inline `DEFAULT_` constants.** Drop both module-level constants;
  inline the `Path(...)` literals at the single `_sync` call sites.
  The `DEFAULT_` prefix promised an alternative form that doesn't
  exist; removing the surface is the deletion route.
- **converge `project_root` docstrings.** Rewrite the `project_root`
  paragraph at six sites onto the `sync_file`/`remove_file` voice.
  Adapted where appropriate: read-side functions drop the absolute-
  path edge case (it doesn't apply); `sync_skill` drops it too (no
  `path` parameter, configured outputs are all relative); functions
  with required `project_root` (`_write_stubs`, `_build_stubs`) drop
  the optional framing.

After this, every site on the sync pipeline names the parameter
`project_root`, every docstring describes it the same way, and no
unused-prefix constant rides the surface.

Closes #99.

### DIFF
--- a/.uncoded/namespace.yaml
+++ b/.uncoded/namespace.yaml
@@ -12,7 +12,6 @@ src/:
     __init__.py:
       __version__:
     cli.py:
-      DEFAULT_MAP_OUTPUT:
       _sync:
       main:
     config.py:
@@ -69,7 +68,6 @@ src/:
       sync_skill:
     stubs.py:
       VALUE_WIDTH_CAP:
-      DEFAULT_STUBS_OUTPUT:
       StubParam:
         name:
         annotation:
--- a/.uncoded/stubs/src/uncoded/cli.pyi
+++ b/.uncoded/stubs/src/uncoded/cli.pyi
@@ -11,10 +11,8 @@ from uncoded.instruction_files import sync_instruction_file
 from uncoded.namespace_map import build_map, render_map
 from uncoded.serena_setup import setup
 from uncoded.skill import sync_skill
-from uncoded.stubs import DEFAULT_STUBS_OUTPUT, _generate_stubs, _write_stubs
+from uncoded.stubs import _generate_stubs, _write_stubs
 from uncoded.sync import sync_file
-
-DEFAULT_MAP_OUTPUT = Path('.uncoded/namespace.yaml')
 
 def _sync(*, start: Path | None, check: bool) -> int:
     """Sync (or verify) the namespace map, stub files, and instruction-file sections."""
--- a/.uncoded/stubs/src/uncoded/extract.pyi
+++ b/.uncoded/stubs/src/uncoded/extract.pyi
@@ -20,7 +20,7 @@ def extract_module(source: str, rel_path: str) -> ModuleInfo:
     """Parse Python source and extract classes, functions, and constants."""
     ...
 
-def iter_source_files(source_root: Path, base: Path | None) -> Iterator[tuple[str, str]]:
+def iter_source_files(source_root: Path, project_root: Path | None) -> Iterator[tuple[str, str]]:
     """Yield (source_text, rel_path) for every parseable Python file in *source_root*."""
     ...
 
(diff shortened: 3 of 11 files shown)


alimanfoo:
**Recommendation:** approve subject to one small thing on the convergence target itself.

## Non-blocking
1. `remove_file` in `src/uncoded/sync.py:57-67` is part of the reference voice this PR converges everyone onto, but its own docstring omits the absolute-path edge case that `sync_file` (`sync.py:30-34`) and `sync_instruction_file` (`instruction_files.py:171-175`) both state ("If ``path`` is absolute, it's used as-is and ``project_root`` has no effect."). All three share identical `project_root / path` semantics, so the omission is a documentation gap, not a behaviour difference. With five docstrings now redirected onto this reference, the reference itself should carry the same closing sentence — otherwise "one voice" still has remove_file as a quiet outlier. Looks like the same edit, just one the PR didn't reach because sync.py wasn't in the diff.

Issue #100, opened 2026-05-07
[claude] _sync's docstring claims "1 on configuration error" without naming any of the three error sites


alimanfoo:
`cli._sync`'s docstring covers its error-return contract in one phrase: \"In apply mode, returns 0 on success or 1 on configuration error.\" The body has three distinct configuration-error paths that produce that 1, each with its own user-facing message and recovery hint:

- `find_pyproject_toml` returns `None` → `\"Error: No pyproject.toml found. Add [tool.uncoded] source-roots to configure.\"`
- `read_source_roots` raises `LookupError` → re-printed as `f\"Error: {e}\"`
- A configured source root isn't a directory → `\"Error: source root {configured} is not a directory. Check [tool.uncoded] source-roots in pyproject.toml.\"`

The three messages were carefully shaped by #85 (recovery hints) and #88 (UX/contract on the source-root case). After #90 task 4 slimmed the project-anchor side of the docstring, the docstring's coverage of the error contract is now visibly out of proportion: the project-anchor side names `start` and `project_root` plainly; the error side says \"1 on configuration error\" and stops.

**Why this is a recurrence.** #85 and #88 are closed prior chips on the same surface (`_sync`'s configuration-error reporting). Each fixed the user-facing symptom — the message, the hint, the contract — without folding the resulting contract into the docstring. The docstring's silence on the three sites is the through-line: a future contributor reading `_sync` learns only that 1 means \"configuration error,\" not which paths produce it or what each one promises the user.

**Suggested direction.** The docstring's second paragraph should name the three error sites at contract level — what each one detects, what it tells the user, what recovery hint it offers — so the next change to any of the three has the docstring as a place to converge against, not a one-phrase summary.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

alimanfoo:
Closing as won't-fix: premise rejected on review of #105.

This issue's premise was that `cli._sync`'s docstring should name the
three configuration-error sites at contract level. The body of `_sync`
is itself the contract for what stderr says: three `print(...)` calls
each followed by `return 1`, with the actual message text inline. A
docstring repeating that creates a parallel surface that has to track
the body on every change, without adding clarity for a reader who can
read the body directly.

The session opened this issue and worked on it (two tasks plus one
review-driven polish task). On PR review, the user pointed out that the
expansion was duplicating what the body already showed. The session
trimmed the docstring back below the pre-session baseline. Net session
[...]

PR #105, opened 2026-05-08
[claude] one project_root contract, one stubs pipeline path

### LINKED ISSUE #97: [claude] cli._sync's missing-pyproject hint points one step past the user's actual state
When a user runs \`uncoded sync\` in a directory with no \`pyproject.toml\`, \`cli._sync\` prints (\`src/uncoded/cli.py:46-52\`):

\`\`\`
Error: No pyproject.toml found. Add [tool.uncoded] source-roots to configure.
\`\`\`

The problem statement is correct — the user has no \`pyproject.toml\`. The recovery hint, however, references \`[tool.uncoded] source-roots\`, which is a section *inside* a \`pyproject.toml\` they don't yet have. To act on the hint they must first create the file, then add the section. The message names step two while skipping step one.

The other two configuration-error sites in \`_sync\` don't have this asymmetry — both fire only when a \`pyproject.toml\` exists and the user just needs to fix something inside it (\"Add [tool.uncoded] source-roots to configure\" for the missing-section case, \"Check [tool.uncoded] source-roots in pyproject.toml\" for the bad-path case). The recovery hint matches the user's actual state in both. Site #1 is the one whose recovery hint reaches one decision past where the user actually is.

The bar this clears: a behaviour gap with a real consumer. The consumer is anyone running \`uncoded sync\` for the first time in a directory that isn't yet a Python project, or who's run it in the wrong directory. The hint they receive doesn't quite recover them from the state they're in.

Prior chips on \`_sync\`'s configuration-error reporting: #85 (KeyError quoting + asymmetric recovery hints), #88 (path leak + idiom asymmetry across the three sites). Neither chip touched site #1's hint targeting in particular — #85 was about quoting at site #2, #88 was about polish at site #3 and the duplicate-walk pattern. Site #1's hint has been stable across both sessions.

Possible direction: a one-line message tweak so the recovery hint matches the user's actual state. Something like \"Create a pyproject.toml with a \[tool.uncoded\] source-roots entry.\" (verbatim phrasing TBD; the substantive change is naming the missing file before the missing section).

### LINKED ISSUE #102: [claude] project_root parameter spans three contracts after #99's name convergence
After #99 collapsed the project-anchor surface onto a single name (`project_root`) and a single docstring voice, the parameter still has three different contracts across the eight functions that take it:

- **Optional, default `None` means "no anchor; use `path` as-is."**  
  `sync_file`, `remove_file`, `sync_instruction_file`, `sync_skill`.
- **Optional, default `None` means "fall back to `Path.cwd()`."**  
  `iter_source_files`, `walk_source` in `src/uncoded/extract.py`.
- **Required `Path`, caller must pre-resolve.**  
  `_write_stubs`, `_build_stubs` in `src/uncoded/stubs.py`.

A reader of `iter_source_files(source_root, project_root=None)` can't predict from the signature alone whether `None` means "no anchor" or "fall back to cwd." They have to read the body. The convergence in #99 named the surface; the contract under it still varies.

Prior chips on this family: #54, #60, #68, #77, #84, #90, #99. Seven over the project-anchor surface. Each previous chip resolved a smaller piece — dormant API, cwd vs project root, writer-side rename, read-side rename, prose voice. The contract variation is what's left after the names line up.

Possible direction: the differences may not be accidental — readers benefit from the cwd fallback for ergonomic callers, internal helpers can require resolved paths because their callers always do that work. But the *unstated* aspect is the gap. Either name the variation in module-level prose, or harmonise the contract (for example, drop the cwd fallback on the readers and require every caller to pass `project_root`). The decision is about what `project_root: Path | None = None` commits to its caller.

### LINKED ISSUE #103: [claude] _build_stubs duplicates _sync's pipeline shape
`src/uncoded/stubs.py::_build_stubs` runs the same `iter_source_files` → `_generate_stubs` → `_write_stubs` chain that `src/uncoded/cli.py::_sync` runs inline. Two parallel paths through the same pipeline.

The duplication has a maintenance cost: any change to the pipeline shape has to land in both places. #99's tasks 1 and 3 demonstrate the cost — the read-side rename and the docstring convergence both updated the call in `_sync` and the call in `_build_stubs`. Future pipeline changes will pay the same overhead.

`_build_stubs` exists as a test-only helper (its docstring says so). #89 chipped at the same surface from a different angle — the function was named `build_stubs` and shaped like public API; #98 task 9 closed it by privatising the name. The structural question underneath wasn't asked: should the test helper duplicate the pipeline at all, or should tests call into shared internals?

Possible direction: extract the three-step pipeline into a single internal helper (`_run_stubs_pipeline(source_root, project_root, ...)` or similar), and have both `_sync` and the tests call it. The test boundary moves from "an end-to-end helper that mirrors `_sync`" to "the same code `_sync` runs, exercised against test inputs." Drift across two paths becomes impossible because there is only one path.

Prior chip on this surface: #89 (closed by #98 task 9 — privatisation).

### LINKED ISSUE #104: [claude] stubs.py imports _property_kind from extract.py across the module boundary
`src/uncoded/stubs.py` imports `_property_kind` from `uncoded.extract` (line 10). The leading underscore advertises a private symbol; the cross-module import reaches across that boundary.

The choice is deliberate — it avoids duplicating the small property-classification logic between the two modules. The cost is structural: a future change to `_property_kind` in `extract.py` silently affects `stubs.py`. The leading underscore tells the maintainer "this is local to `extract.py`," but it isn't.

The repo's own coherence-review skill (generated by `uncoded sync`) treats this kind of pattern — `from module import _thing` — as a structural symptom worth flagging.

Possible direction: three options.
- **Promote.** Rename `_property_kind` to `property_kind` in `extract.py`. Acknowledges that it's part of the module's intended interface for in-package consumers.
- **Move.** Relocate the function to a shared utility module so both `stubs.py` and `extract.py` import it without crossing a private boundary.
- **Document.** Keep the underscore but add a comment naming `stubs.py` as an in-package consumer. Cheapest, but doesn't change the structural fragility.

The shape of the answer depends on whether `_property_kind` belongs to `extract.py`'s domain (parsing) or to a domain shared with `stubs.py` (decorator classification).

alimanfoo:
`project_root` was a parameter on eight functions across the sync
pipeline, with three different contracts under one name:

- `Path | None = None`, where `None` meant "no anchor; use the path
  as-is" (`sync_file`, `remove_file`, `sync_instruction_file`,
  `sync_skill`).
- `Path | None = None`, where `None` meant "fall back to
  `Path.cwd()`" (`iter_source_files`, `walk_source`).
- Required `Path` (`_write_stubs`, `_build_stubs`).

A reader of `iter_source_files(source_root, project_root=None)` had
no way to predict from the signature which `None` meant which. And
no caller exercised the optional variants: `cli._sync` always passes
a resolved `project_root`, and every test passes `tmp_path`. The
optionality was speculative.

This PR collapses the three contracts into one: `project_root:
Path`, required everywhere.

Three sibling concerns in the same pipeline travelled with it:

- The stubs pipeline (`iter_source_files` → `_generate_stubs` →
  `_write_stubs`) ran in two places: `cli._sync`'s loop and the
  test-only `_build_stubs`. They've collapsed into a single shared
  entry point, `build_stubs`, taking pre-iterated `files` so `_sync`
  doesn't double-walk source.
- `_property_kind` and `_build_stubs` were imported across the
  module boundary with leading underscores — a pattern the project's
  own coherence-review skill flags. Both are intentionally shared
  in-package, so they've been promoted (`property_kind`,
  `build_stubs`).
- `cli._sync`'s missing-pyproject hint pointed at a config section
  inside a file the user didn't have yet. It now names the missing
  file first.

Closes #97
Closes #102
Closes #103
Closes #104

#100 is not closed by this PR — its premise (that the docstring
should name the three configuration-error sites at contract level)
was rejected on review. The body of `_sync` is the contract for what
stderr says; the docstring expansion duplicated it. The docstring is
shorter than its pre-session state.


alimanfoo:
**Recommendation:** approve subject to a couple of prose and ergonomics nits.

## Non-blocking
1. **`_sync` docstring (`src/uncoded/cli.py`, lines 31–39) is dense on first read.** The paragraph opens with a forward reference ("the three sites are checked before the apply/check branch") and then runs three independent error paths together as one block of "If … If … If …". The reader has to count the conditions to find the "three" referenced earlier. Splitting into a short lead ("Three configuration errors are checked before the apply/check branch and return 1 in both modes:") followed by one bullet or sentence per case would let each rule stand alone. Mentioning `LookupError` and `read_source_roots` in the middle case also leaks the implementation; describing the user-facing message would read more naturally for someone reading the contract.

2. **`iter_source_files` docstring's consumer list is now incomplete (`src/uncoded/extract.py`, lines 109–112).** The trailing sentence still names "downstream consumers (`walk_source`, `_generate_stubs`)" as if exhaustive. After this PR, `cli._sync` consumes the iterator's output directly and feeds it to `extract_modules` (and to `build_stubs`). Either reword to read as illustrative ("e.g., `extract_modules`, `_generate_stubs`") or list the actual consumers — `extract_modules` and `_generate_stubs` are the functions that rely on the parseable-source guarantee.

## Nits
1. **Test boilerplate in `tests/test_stubs.py`.** Every `TestBuildStubs` / `TestBuildStubsCheckMode` call now repeats `files=list(iter_source_files(src, project_root=tmp_path))` (~20 sites). Consider folding the iteration into the `_setup` helper or a tiny per-test wrapper so the call sites stay focused on what each test asserts. Pure ergonomics — semantics unchanged.

2. **Implicit cross-call `project_root` precondition for `build_stubs` (`src/uncoded/stubs.py`, line 444).** The two-step API now requires callers to pass the same `project_root` to both `iter_source_files` (for relativising) and `build_stubs` (for writing). The docstring says "source paths in `files` are already relative to it" but stops short of saying the two values must match. Worth stating outright so a future caller doesn't pass mismatched anchors and silently produce wrong-looking stub paths.

src/uncoded/cli.py
```diff
@@ -28,10 +28,20 @@ def _sync(*, start: Path | None = None, check: bool = False) -> int:
...
+    entry. If the located ``pyproject.toml`` has no ``[tool.uncoded]``
+    ``source-roots`` entry, stderr names the missing section and tells
+    the user to add it. If a configured source root is not a directory,
+    stderr names the entry as the user wrote it and points at the
+    ``[tool.uncoded]`` ``source-roots`` key.
+
+    After configuration validates, ``check=True`` reports each prospective
+    write without mutating the on-disk tree, returning 1 if any step would
+    write (so CI can gate on a stale index) and 0 if the tree is already
+    in sync. Apply mode performs those writes and returns 0.
```
alimanfoo: Is this all really needed? It feels like a lot of this is just describing code. Is the code not self-explanatory?

Issue #95, opened 2026-05-05
Agent-navigation prose lives in two parallel templates that drift


alimanfoo:
The orient/understand/act walkthrough — the dispatch rule, the namespace map step, the `.pyi` stub step, the Serena tool list — exists in two near-identical copies:

- `src/uncoded/instruction_files.py` (the `_SECTION_BODY`-style template injected into `CLAUDE.md` and `AGENTS.md`)
- `src/uncoded/skill.py` (the `_SKILL_CONTENT` template for the coherence-review skill)

There is no shared source for the parts that are identical, and no test that locks the parts that should agree. Edits intended to apply uniformly to "the agent navigation guidance" must be made in both places by hand. #92 / #93 was an instance: the misleading "in source order" claim lived in both templates, so the fix had to edit both to remove the same sentence from rendered prose.

The behaviour gap is real even though the two templates may legitimately need to differ for their respective audiences (a session-start instruction vs a skill-invoked prompt). The gap is silent drift: an edit to one copy that should also have hit the other will not produce a test failure or a check-time warning. The next agent landing on the dispatch-rule prose will be the third agent fixing the same sentence in three places (`namespace_map.py`, `instruction_files.py`, `skill.py` — exactly what #93 did).

Possible directions, listed for the issue's body, not as a decision:

- **Single source.** Factor the identical fragments out into a shared constant (or fragment file) imported by both templates. Audience-specific framing wraps the shared core.
- **Lockstep test.** Keep the two templates separate, but add a test that asserts the dispatch-rule sentence and the orient/understand/act bullets are byte-identical (or normalised-identical) across `_SECTION_BODY` and `_SKILL_CONTENT`. Fails loudly on accidental drift.
- **Explicit divergence registry.** A small allowlist that names the parts of the two templates that are deliberately different, leaving the rest pinned to match.

The contract being defended is "agent-facing navigation guidance is consistent across the surfaces an agent encounters it on." The recurrence pattern (#92 hitting the same surface) is itself the evidence that the contract is currently held only in author-discipline.

Surfaced in the post-merge sweep for #93 by the reviewer; independently observed in the same form by the maintainer.

Prior chip on the same surface: #92 (the dropped "source order" claim, edited in both templates).

alimanfoo:
Closing as overstated on inspection.

[...]

PR #108, opened 2026-05-11
[claude] drop redundant tests and dead public API (#94, #106, #107)

### LINKED ISSUE #94: Renderer's within-file emission order is enforced by tests but not documented
The namespace map renderer (`build_map` in `src/uncoded/namespace_map.py`) emits within-file symbols in a specific order:

1. **Module-level constants** first
2. **Classes** next, with **attributes before methods**
3. **Functions** last
4. Within each list, **insertion order** as supplied to `ModuleInfo`

Three tests pin this contract:

- `tests/test_namespace_map.py::TestBuildMap::test_constants_precede_classes_and_functions` — pins constants ahead of classes and functions
- `tests/test_namespace_map.py::TestBuildMap::test_class_and_function_insertion_order_preserved` — pins classes-then-functions and within-list insertion order
- `tests/test_namespace_map.py::TestBuildMap::test_class_with_attributes_and_methods` — pins attributes-before-methods within a class (with an inline comment, no named test)

But no surface a consumer actually reads says any of this:

- `build_map`'s docstring (`src/uncoded/namespace_map.py:35-71`) describes only the directory/file structure
- The `namespace.yaml` HEADER (`src/uncoded/namespace_map.py:11-18`) used to claim "in source order" — that claim was removed in #92 / #93 because it was misleading. Nothing replaced it.
- The agent-facing prose in `src/uncoded/instruction_files.py` and `src/uncoded/skill.py` describes "directories, files, classes, methods, functions" without naming the within-file tier order

This is the same shape #92 had: a contract enforced somewhere, invisible at the surface where its consumer reads. #92 was resolved by removing the (misleading) prose; the tests still demand the order, but now silently. An agent reading `namespace.yaml` to plan navigation has no way to know which order to expect.

Two ways to close the gap:

1. **Document the order.** Name it once in `build_map`'s docstring and once in the namespace.yaml HEADER (and possibly the agent prose). The tests then defend a named contract, not a tacit one.
2. **Drop the tests.** If the order has no consumer that depends on it, the three tests are pinning surface — drop them, and the renderer is free to emit symbols in whatever order is convenient.

The decision turns on whether any consumer actually relies on tier order. The session that surfaced this finding (#93) leaned toward "order is decorative for the agent" during scoping, but the tests have not been reconciled with that framing.

Surfaced in the post-merge sweep for #93 by both the maintainer and reviewer roles, independently, in the same shape.

### LINKED ISSUE #106: [claude] walk_source is test-only public API after #105
After #105 collapsed the stubs pipeline, `cli._sync` builds modules
inline via `iter_source_files` + `extract_modules` rather than going
through `walk_source`. The function now has zero production callers;
only `tests/test_extract.py` (`TestWalkSource` plus one `TestExtractModule`
case) exercises it.

The same shape was filed and closed at #89 for `build_stubs` — "shaped
like public API but only tests call it." #89 closed by privatising to
`_build_stubs`. In #105 we promoted `_build_stubs` back because it
gained a production caller (`cli._sync` now calls it). `walk_source`
did the opposite: it lost a production caller without gaining a
different one.

Suggested direction (the deciding session can pick):

- **Delete.** `walk_source`'s body is a one-line wrapper around
  `extract_modules(iter_source_files(source_root, project_root))`. The
  six test call sites would inline that, or use a small test-local
  helper if the repetition matters.
- **Privatise.** Rename to `_walk_source`. Tests already import directly,
  so the cost is small.

Anyone navigating the namespace map sees
`walk_source` exposed as part of the public extraction surface, but
production doesn't use it.

### LINKED ISSUE #107: [claude] generate_section duplicates SECTION as redundant public surface
`src/uncoded/instruction_files.py` defines two public symbols carrying
the same value:

- `SECTION` — module-level f-string built at import time.
- `generate_section()` — parameterless function whose entire body is
  `return SECTION`.

Redundant public surface. A public constant and a public parameterless
function in the same module where the function's sole body is `return
<constant>`. Both symbols being public exposes an implementation
detail unnecessarily — only one needs to be public.

Production caller: `sync_instruction_file` calls `generate_section()`
once. Test callers: six methods in `tests/test_instruction_files.py`.

Suggested direction: drop `generate_section`; have callers reference
`SECTION` directly. The production call and the six tests would change
to reference the constant.

alimanfoo:
Removes three surfaces that defended or exposed nothing a consumer reads. (A fourth issue scoped alongside these, #95, was closed separately as overstated on inspection — see the comment on that issue.)

Closes #94. Closes #106. Closes #107.

## Changes

**#94 — within-file tier-order tests in `TestBuildMap`.** Three tests in `tests/test_namespace_map.py` pinned the order in which `build_map` emits within-file symbols (constants → classes (attributes-then-methods) → functions). The agent reads the namespace map in full at session start, so within-file order is decorative — #93 already removed the "in source order" claim from the header for this reason. The tests pinned a contract with no consumer; dropping them frees the renderer to emit symbols in whatever order is convenient without a test alarm. `build_map`'s actual iteration order is unchanged.

An order-agnostic dict-equality test was kept (commit `5341453`) to cover the real behaviour the deleted `test_class_with_attributes_and_methods` carried alongside the order claim: that `ClassInfo.attributes` and `ClassInfo.methods` both flow into the file entry.

**#106 — `walk_source` deleted.** `walk_source` was a one-line wrapper around `extract_modules(iter_source_files(...))`. After #105 collapsed the stubs pipeline, `cli._sync` calls the composition directly and `walk_source` had zero production callers — only six test callers. The test sites now call `extract_modules(iter_source_files(...))` inline. `TestWalkSource` was renamed to `TestIterAndExtract`. Commit `cfd5799` moved `test_module_with_only_constants_is_kept` to `TestExtractModules` — the renamed class made visible that the test was placed by its old caller, not by what it actually exercises (`extract_modules`'s skip-empty filter).

**#107 — `generate_section` deleted.** `generate_section` was a parameterless function whose body was `return SECTION`. Both the function and the constant were public; only one needed to be. `sync_instruction_file` and the eight test call sites now reference `SECTION` directly. `TestGenerateSection` was renamed to `TestSection`.


alimanfoo:
**Recommendation:** approve subject to one stale docstring reference.

## Non-blocking
1. `src/uncoded/stubs.py:463` — the docstring on `build_stubs` still cites `:func:`walk_source`` when explaining how stub paths align with the namespace map. The PR deletes `walk_source`, so this cross-reference now points at nothing. It's the same edit the PR is already making, just on a surface the task list didn't name. Reword to cite `iter_source_files` (which now produces the rel_path the docstring is describing) or drop the parenthetical.

## Nits
1. `tests/test_extract.py` — the renamed `TestIterAndExtract` class keeps method names from the old `TestWalkSource` shape (`test_basic_walk`, `test_nested_subpackage`, `test_includes_init_with_symbols`, `test_skips_empty_init`, `test_skips_syntax_errors`). They still describe the filesystem-walk half of the composition, so they're not wrong — just a touch stale next to the new class name. Worth a pass if you're already renaming.

Issue #57, opened 2026-04-30
[claude] uncoded's index doesn't capture module docstrings — a class of drift escapes coherence-review


alimanfoo:
`src/uncoded/stubs.py::extract_stub` captures function and class docstrings (first sentence) but not module docstrings. `src/uncoded/namespace_map.py::build_map` doesn't carry docstrings at all.

So when a module docstring drifts away from the module's actual contents — e.g. the `config.py` drift fixed in #49 / PR #53, where the docstring said `"""Read uncoded configuration from pyproject.toml."""` but the module had grown a function reading PEP 621 project metadata — uncoded's own static index can't help an agent catch it. The drift had to be found and fixed by hand.

Given the project's own framing ("agents fill gaps from pretraining rather than reading the actual code, ... a static, pre-computed index gives agents a top-down view"), this is a notable gap in uncoded's coverage of its own value proposition. Capturing module docstrings in `extract_stub` (first sentence, mirroring the existing pattern for functions and classes) and surfacing them in the namespace map would let coherence-review reason about module-level documentation drift the same way it can reason about function-level drift.

Surfaced by the developer agent in the post-merge sweep for PR #53.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

alimanfoo:
Duplicate of #63. Closing — first batch of `gh issue create` calls reported an internal tool error so I retried, but the originals had silently succeeded. #63 is the canonical version of this concern.

PR #112, opened 2026-05-11
[claude] drop docstring excerpts from stubs (#110)

### LINKED ISSUE #110: [claude] stub-based index hides docstring drift past the first sentence
`.pyi` stubs generated by `src/uncoded/stubs.py::extract_stub` carry only the **first sentence** of each function and class docstring. Anything in subsequent paragraphs — cross-references (`:func:`name``), worked examples, parameter notes, behavioural caveats — is absent from the index. Agents reading stubs first (which uncoded's own dispatch rule recommends) see no signal that the deeper docstring exists, let alone that it might be stale.

PR #108 illustrated the gap concretely. `build_stubs`'s third paragraph in `src/uncoded/stubs.py` referenced `:func:`walk_source`` to ground its `rel_path` claim. `walk_source` was deleted earlier in the same PR; the dangling reference survived through #105 and #106 because no one with their eyes on the stub had reason to look at the third paragraph. The reviewer agent caught it on a fresh diff read, not via the index. The same gap would have been there if it had been a renamed symbol rather than a deleted one, and across as many sessions as it took someone to land on the body of that docstring.

Related but distinct from #57 / #63 (both closed by adding module docstrings to stubs). Those issues established that the index should cover docstrings; this one is about how much of each docstring the index covers.

Possible directions (the deciding session can pick):

- **Carry more of each docstring in the stub.** Full docstring rather than first sentence, with the cost of a larger index.
- **Lint cross-references at sync time.** When `uncoded sync` regenerates the index, parse `:func:`/`:class:`/`:meth:`/`:attr:` markers in source docstrings and verify each target still exists. Fail or warn on misses.
- **Update the dispatch rule.** Remind agents that source `.py` reads — not stub reads — are the right surface when renaming or deleting a symbol, since the stub only carries the first-sentence docstring view.

The concern is the gap in the index's value proposition: uncoded's own framing is that the index gives agents a top-down view of the codebase, and docstring bodies are part of what that view should cover for symbols whose contracts include cross-references.

Surfaced in the post-merge sweep for PR #108 by the developer agent.

alimanfoo:
Closes #110.

The stub-based index carried only the first sentence of each docstring, via a 37-line `_first_sentence` heuristic. Anything past the first sentence — cross-references, worked examples, parameter notes, behavioural caveats — sat outside the index. PR #108 illustrated the gap concretely: a dangling `:func:` reference to a deleted symbol survived three earlier PRs in the third paragraph of a docstring no stub-reader had reason to look at.

The first-sentence machinery had accumulated five issues' worth of fixes (#57, #63, #72, #81, #86). Each chipped a different symptom of the same underlying contract: partial-docstring extraction is a heuristic-driven projection of fuller prose. Rather than widen the projection or add scaffolding around it, this drops the projection entirely. Stubs now carry signatures and structure — the things a stub is good at — and agents read source via Serena's `find_symbol(include_body=True)` when they want the prose.

The skill that runs the coherence-review sweep is updated to match: names and signatures still come from the stub; docstrings come from `find_symbol`. The "first-sentence docstrings" line in README, AGENTS, CLAUDE, and the synced section template is updated to reflect the new contract. A fossil blank line in the class-header render path, exposed once every class lost its docstring, is also removed.


alimanfoo:
**Recommendation:** approve subject to a couple of small things.

## Non-blocking

1. Classes whose only body is a docstring now render to an invalid `.pyi` stub: just `class Foo:` with no body. In `render_stub` (`src/uncoded/stubs.py` ~284–295), the class header is emitted, the docstring rendering is gone, and the unconditional `lines.append("")` that used to sit there is gone too — for a class with no attributes and no methods, nothing more is appended. The trailing `rstrip()` then eats any final newline, leaving a header with no body. The repo's own classes all have attributes or methods, so dogfood stubs are clean; downstream users with marker classes (`class Sentinel: """A marker."""`) would now get malformed stubs. `tests/test_stubs.py::test_class_no_bases` exercises an empty-body class but only asserts on `.bases`, so the regression isn't caught. A `pass` or `...` line in the no-attrs-no-methods branch of `render_stub` would close it.

2. The new promissory-sweep workflow text reads awkwardly: in `src/uncoded/skill.py` and the generated `SKILL.md` files (`Step 3: Promissory sweep`), the prose says *"For each non-trivial public symbol... read the stub for name and signature, then call Serena's `find_symbol` with `include_body=True` for the docstring."* The stub is per-file, not per-symbol, so the per-symbol framing suggests a stub read per symbol. And `find_symbol(include_body=True)` returns the full symbol body, not just the docstring — the later "When to read further" paragraph acknowledges this, but the lead sentence misleads. Worth tightening: read the stub once for the file's names and signatures; call `find_symbol(include_body=True)` per symbol when you need the docstring (and incidentally get the body).

## Nits

1. `README.md:127` still describes the split as *"uncoded provides a stable map and semantic summary; Serena resolves the current source body."* The "semantic summary" half was the first-sentence docstrings the PR removes. With stubs now signature-and-structure only, a phrase like "stable map and signature index" would match the current contract more closely.

PR #114, opened 2026-05-12
[claude] tighten render_stub tests, rename TestExtractModule(s)

### LINKED ISSUE #109: [claude] TestExtractModule and TestExtractModules read as a typo
`tests/test_extract.py` defines two top-level test classes that differ only by a trailing `s`: `TestExtractModule` (covering `extract_module(source, rel_path)`, the single-module extractor) and `TestExtractModules` (covering `extract_modules(files)`, the iterable variant). The names map to their subjects accurately — neither is wrong — but the one-letter difference creates friction for both human and agent readers. Developer-agent feedback during PR #108 tasks 3 and 6 noted having to read the trailing `s` carefully each time to keep the classes straight; the friction was real during work that moved a test between them.

Suggested directions (the deciding session can pick):

- **Rename one for daylight.** Something like `TestExtractModuleFromSource` / `TestExtractModulesFromFiles`, or another pair that does not depend on a single character.
- **Fold the plural class into `TestIterAndExtract`.** All four `TestExtractModules` tests work over multi-file lists; `TestIterAndExtract` already covers the iter+extract composition. Merging eliminates the visual collision by removing the plural class.

The risk in leaving it as-is: tasks moving tests between the two will keep paying the cost, and the names are easy to mis-cite in PR descriptions, commit messages, or audit reports.

Surfaced in the post-merge sweep for PR #108 by the developer agent.

### LINKED ISSUE #113: [claude] render_stub tests defend substrings, not Python validity
`render_stub` in `src/uncoded/stubs.py` produces `.pyi` stubs that agents read as Python. The tests under `tests/test_stubs.py::TestRenderStub`, and the empty-class case in `tests/test_stubs.py::TestExtractStub::test_class_no_bases`, defend that output loosely — almost all `TestRenderStub` assertions are substring `in` checks, and `test_class_no_bases` only asserts on `.bases` without rendering at all. None of them defend the contract a consumer actually cares about: the rendered stub is valid Python.

The marker-class regression Ada flagged on PR #112 illustrated the gap. A class with only a docstring rendered to `class Foo:` with no body — invalid `.pyi` — and the existing tests didn't notice. `test_class_no_bases` uses `class Plain: pass` but never reaches the rendered output. `TestRenderStub`'s checks (`"def greet(" in output`, `"class Worker:" in output`) would pass on any output containing those substrings anywhere, regardless of surrounding layout.

The new test added in this PR — `tests/test_stubs.py::TestRenderStub::test_class_with_no_members_renders_body` — is the tighter shape: a precise substring (`"class Sentinel:\n    ...\n"`) plus a `compile()` smoke check confirming the rendered text parses as Python. The same shape would catch the same class of bug in other parts of the renderer.

Possible directions:

- **Tighten existing render tests.** Replace loose substring `in` checks with precise sub-strings that pin layout, and add a `compile()` smoke check to every test that emits a full rendered stub.
- **One render-validity smoke test.** A single test that takes a representative source file, runs `extract_stub` → `render_stub`, and asserts `compile(result, ...)` succeeds. Cheap; catches a wide class of layout bug.
- **Both.** The two defend different things: precise sub-strings pin the shape; `compile()` smokes the result.

Surfaced in the post-merge sweep for PR #112 by the reviewer and the developer, independently.

alimanfoo:
Two maintenance fixes from post-merge sweeps on PRs #108 and #112.

## TestExtractModule vs TestExtractModules (#109)

The two top-level test classes in `tests/test_extract.py` differed only by a trailing `s`, costing reading time during work that moved tests between them. Renamed to `TestExtractModuleFromSource` and `TestExtractModulesFromFiles` — the layered design (single-source transform, filesystem composition, plural transform) is preserved.

Closes #109.

## render_stub test assertions (#113)

`TestRenderStub` defended its output with substring `in` checks that would pass on missing-body, missing-newline, or wrong-indent output — the marker-class bug fixed in #112 shipped because nothing exercised `compile()` on a rendered stub. Two changes:

- Each `TestRenderStub` assertion now pins layout — anchored on newlines and explicit indentation, so a regression in `render_stub` would localise to the affected test.
- A new test, `test_renders_valid_python_for_representative_source`, drives a representative source through `extract_stub` → `render_stub` and asserts the result parses via `compile()`. The source covers every branch in the renderer, including the no-members class branch that prompted #112.

`test_class_no_bases` was consolidated into `test_class_with_no_members_renders_body` — after #112 they tested the same case.

Closes #113.

### DIFF
--- a/.uncoded/namespace.yaml
+++ b/.uncoded/namespace.yaml
@@ -297,6 +297,7 @@ tests/:
       test_constant_bare_annotation_rendered:
       test_type_alias_pep695_rendered:
       test_unannotated_class_attribute_rendered:
+      test_renders_valid_python_for_representative_source:
     TestBuildStubs:
       _setup:
       _build:
--- a/.uncoded/namespace.yaml
+++ b/.uncoded/namespace.yaml
@@ -286,7 +286,6 @@ tests/:
       test_async_function_prefix:
       test_function_with_annotations:
       test_class_with_bases:
-      test_class_no_bases:
       test_class_with_no_members_renders_body:
       test_attribute_with_annotation:
       test_method_indented:
--- a/.uncoded/namespace.yaml
+++ b/.uncoded/namespace.yaml
@@ -155,7 +155,7 @@ tests/:
       test_reads_configured_list:
       test_empty_list_is_respected:
   test_extract.py:
-    TestExtractModule:
+    TestExtractModuleFromSource:
       test_classes_and_functions:
       test_async_functions_and_methods:
       test_empty_module:
@@ -174,7 +174,7 @@ tests/:
       test_includes_init_with_symbols:
       test_skips_empty_init:
       test_skips_syntax_errors:
-    TestExtractModules:
+    TestExtractModulesFromFiles:
       test_returns_module_info_per_parseable_file:
       test_preserves_source_order:
       test_module_with_only_constants_is_kept:
--- a/.uncoded/stubs/tests/test_extract.pyi
+++ b/.uncoded/stubs/tests/test_extract.pyi
@@ -3,7 +3,7 @@
 import textwrap
 from uncoded.extract import extract_module, extract_modules, iter_source_files
 
-class TestExtractModule:
+class TestExtractModuleFromSource:
     def test_classes_and_functions(self):
         ...
 
@@ -56,7 +56,7 @@ class TestIterAndExtract:
     def test_skips_syntax_errors(self, tmp_path, capsys):
         ...
 
-class TestExtractModules:
+class TestExtractModulesFromFiles:
     def test_returns_module_info_per_parseable_file(self):
         ...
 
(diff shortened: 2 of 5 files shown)


alimanfoo:
**Recommendation:** looks good, one non-blocking finding.

## Non-blocking
1. `tests/test_stubs.py::TestRenderStub::test_rendered_stub_has_no_line_range_comments` (around tests/test_stubs.py:300-322) still uses the loose substring style this PR is tightening — `"VALUE = 1" in output`, `"class Worker:" in output`, `"    def work(self):" in output`. GH113 cited `"class Worker:" in output` as the exemplar of the pattern the PR is trying to remove, and that assertion sits in this test. The primary check here is `"# L" not in output`, but the surrounding sanity checks would still pass on a layout regression that the rest of the suite would now catch. Worth bringing in line with the other `TestRenderStub` assertions so the tightening is uniform.

PR #117, opened 2026-05-12
[claude] tighten render_stub tests, rename is_pep695_alias

### LINKED ISSUE #115: [claude] TestRenderStub tightening missed test_imports_rendered and varargs/kwargs
PR #114 tightened TestRenderStub assertions in `tests/test_stubs.py` to anchor on newlines and explicit indent, and added a compile() smoke test over a representative source. Two adjacencies on the same surface were missed.

1. `tests/test_stubs.py::TestRenderStub::test_imports_rendered` still uses `"import os\nfrom pathlib import Path" in output` — no trailing-newline anchor. Every other module-level pin in the class is now newline-anchored. The imports branch of `render_stub` was overlooked because PR #114's brief framed the issue around the constants/functions/classes cases that motivated PR #112.

2. The smoke source in `test_renders_valid_python_for_representative_source` (added by PR #114) covers every other branch in `render_stub` but skips varargs and kwargs. `_render_param` in `src/uncoded/stubs.py` handles `*` and `**` prefixes; a layout regression there is currently caught by neither the precise tests (none defend varargs) nor the compile smoke.

Suggested direction:
- Tighten the imports assertion to `"import os\nfrom pathlib import Path\n" in output` to match the sibling pins.
- Add a function like `def build(*args: str, **kwargs: int) -> None: pass` to the smoke source.

Noticed after PR #114 merged, by the developer, the maintainer, and the reviewer.

### LINKED ISSUE #116: [claude] StubAssignment.is_type_alias name doesn't match its semantics
`StubAssignment.is_type_alias` in `src/uncoded/stubs.py` is True only for PEP 695 aliases (`type X = int`) and False for classic-style aliases (`X: TypeAlias = int`) — even though both are type aliases. A reader looking at the field name reasonably assumes both forms get the flag; only the PEP 695 form does. `tests/test_stubs.py::TestExtractStub::test_type_alias_classic` asserts exactly the asymmetric behaviour (`is_type_alias is False` for the classic case), but the asymmetry is implicit; a reader has to walk through the test to discover it.

The flag's actual job is narrower than its name. It tells `_format_assignment_body` whether to emit the `type` prefix. The classic case carries the `TypeAlias` distinction in its `annotation` field and is rendered through the standard `name: annotation = value` path.

Closed #70 cited the same flag on a different concern (namespace-map rendering of type aliases vs constants); that one was closed wontfix. This is the field-naming/semantics concern.

Suggested directions:
- Rename the flag to match its actual meaning, e.g. `is_pep695_alias`.
- Replace the bool with an enum capturing the three assignment shapes explicitly — plain, PEP 695 alias, classic TypeAlias — and dispatch the renderer on the enum.
- Keep the name and add a docstring on the field explaining the PEP 695 specialisation.

Noticed after PR #114 merged, by the maintainer.

alimanfoo:
Two small clean-ups on the stub rendering surface.

PR #114 tightened TestRenderStub assertions but two cases slipped through. `test_imports_rendered` still asserts a substring without the trailing-newline anchor every other module-level pin in the class uses. The smoke source in `test_renders_valid_python_for_representative_source` covers every other rendering branch but skips varargs (`*args`) and kwargs (`**kwargs`). Both are closed here.

`StubAssignment.is_type_alias` is True only for PEP 695 aliases (`type X = int`). Classic `X: TypeAlias = int` carries the alias-ness in its `annotation` field and renders through the standard `name: annotation = value` path. The flag's actual job is narrower than its name — it signals PEP 695-ness so `_format_assignment_body` emits the `type` keyword prefix. The new name matches the semantics.

The redundant `is_type_alias=False` kwarg in `_property_attribute` was dropped at the same time. The dataclass default was already False. After the rename, the kwarg reads as noise: a property is by construction not a PEP 695 alias.

Closes #115
Closes #116

### DIFF
--- a/.uncoded/namespace.yaml
+++ b/.uncoded/namespace.yaml
@@ -78,7 +78,7 @@ src/:
         name:
         annotation:
         value_source:
-        is_type_alias:
+        is_pep695_alias:
       StubClass:
         name:
         bases:
--- a/.uncoded/stubs/src/uncoded/stubs.pyi
+++ b/.uncoded/stubs/src/uncoded/stubs.pyi
@@ -68,7 +68,7 @@ class StubAssignment:
     name: str
     annotation: str | None = None
     value_source: str | None = None
-    is_type_alias: bool = False
+    is_pep695_alias: bool = False
 
 class StubClass:
     name: str
--- a/src/uncoded/stubs.py
+++ b/src/uncoded/stubs.py
@@ -43,7 +43,7 @@ class StubAssignment:
     name: str
     annotation: str | None = None
     value_source: str | None = None
-    is_type_alias: bool = False
+    is_pep695_alias: bool = False
 
 
 @dataclass
@@ -124,7 +124,7 @@ def _extract_assignment(
         return StubAssignment(
             name=node.name.id,
             value_source=_render_value(node.value),
-            is_type_alias=True,
+            is_pep695_alias=True,
         )
 
     if isinstance(node, ast.AnnAssign):
@@ -167,7 +167,6 @@ def _property_attribute(
         name=node.name,
         annotation=ast.unparse(node.returns) if node.returns else None,
         value_source=None,
-        is_type_alias=False,
     )
 
 
@@ -250,7 +249,7 @@ def _render_function(func: StubFunction, indent: str = "") -> list[str]:
 
 def _format_assignment_body(a: StubAssignment) -> str:
     """Render the 'name [: type] [= value]' portion of an assignment."""
-    if a.is_type_alias:
+    if a.is_pep695_alias:
         return f"type {a.name} = {a.value_source}"
     head = f"{a.name}: {a.annotation}" if a.annotation else a.name
     if a.value_source is None:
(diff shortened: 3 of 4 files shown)


alimanfoo:
**Recommendation:** looks good.

The PR closes both follow-ups cleanly. The rename is complete across `src/`, `tests/`, the stub, and `.uncoded/namespace.yaml`, with no stale `is_type_alias` references left. The newline anchor in `test_imports_rendered` matches the sibling pins. The smoke source now exercises `*args` and `**kwargs` through `_render_param`. Dropping the `is_type_alias=False` kwarg from `_property_attribute` is a sensible adjacent cleanup once the name no longer needs the explicit "this is not a type alias" hint.

## Nits
1. `StubAssignment`'s class docstring (`src/uncoded/stubs.py:36-41`) documents the `value_source` tri-state but says nothing about `is_pep695_alias`. The rename does most of the work GH116 asked for, so this is optional — but a one-line note that classic `X: TypeAlias = int` carries its alias-ness in `annotation` instead would pre-empt the same "why is this asymmetric?" question a reader of `test_type_alias_classic` still has to walk through the test to answer.

PR #121, opened 2026-05-13
[claude] tighten _render_param coverage, drop redundant kwarg (#118, #119)

### LINKED ISSUE #118: [claude] _render_param branches keep slipping past TestRenderStub coverage
Every branch in `_render_param` should be defended by TestRenderStub at two levels: the `compile()` smoke source in `test_renders_valid_python_for_representative_source` plus a focused precise-pin test. After #113 (added the smoke source) and #115 (extended it with varargs/kwargs), three branches still slip past:

- **Varargs (`*args`):** the smoke source covers it (added in #117) but there is no focused pin. `compile()` alone won't catch a `*`-prefix drop — `def f(args)` is valid Python, and the function semantics silently change from varargs to required positional.
- **Kwargs (`**kwargs`):** same as varargs.
- **Positional-only (`/`) and keyword-only (`*`) separators:** neither smoke source coverage nor focused pin. `_extract_params` in `src/uncoded/stubs.py` inserts synthetic `StubParam(name="/")` and `StubParam(name="*")`; `_render_param` short-circuits both.

The pattern: each pass closes the branches it sees, but no test names the contract, so further branches keep showing up as gaps.

Suggested direction: name the contract once — every `_render_param` branch is defended by both compile smoke and a focused pin — and enforce it on the remaining branches.

### LINKED ISSUE #119: [claude] _property_attribute passes a redundant value_source=None kwarg
In `src/uncoded/stubs.py`, `_property_attribute` constructs `StubAssignment(name=..., annotation=..., value_source=None)`. `value_source` defaults to `None` on the dataclass, so the explicit kwarg asserts a default the field already guarantees — the same pattern as `is_type_alias=False`, which #117 dropped while closing #116.

After that rename, dropping the `is_type_alias=False` kwarg made the call site read as "name and annotation are all this needs". The remaining `value_source=None` kwarg sits in the same shape: an explicit default with no semantic content for the reader.

Suggested direction: drop the `value_source=None` kwarg.

alimanfoo:
Closes #118.
Closes #119.

## Why

#118 reports that `_render_param` keeps gaining input shapes (varargs, kwargs, the `/` and `*` separators) without the tests catching the gaps. Each prior pass closed the shapes it noticed and moved on. The fix is to name the contract once and enforce it on the remaining shapes, so the next added shape is visible if it has no test row.

#119 is the same kind of redundant-default kwarg as #116 — `_property_attribute` was the last `StubAssignment(...)` call site asserting a field default the dataclass already guarantees.

## What changes

- `TestRenderStub` now has `test_render_param_covers_input_kind`, a parametrised test with one row per shape `_render_param` can receive: regular positional (bare and annotated), varargs (bare and annotated), kwargs (bare and annotated), positional-only `/`, keyword-only `*`. A new shape without a row is visible.
- The representative smoke source in `test_renders_valid_python_for_representative_source` now exercises the four shapes it was missing at compile level: bare `*args`, bare `**kwargs`, `/`, and `*`.
- A separate focused test pins `_render_function`'s `-> {return_annotation}` rendering — restored after the replaced `test_function_with_annotations` was doing double duty.
- `_property_attribute` no longer passes `value_source=None`.

### DIFF
--- a/.uncoded/namespace.yaml
+++ b/.uncoded/namespace.yaml
@@ -284,7 +284,7 @@ tests/:
       test_imports_rendered:
       test_rendered_stub_has_no_line_range_comments:
       test_async_function_prefix:
-      test_function_with_annotations:
+      test_render_param_covers_input_kind:
       test_class_with_bases:
       test_class_with_no_members_renders_body:
       test_attribute_with_annotation:
--- a/.uncoded/namespace.yaml
+++ b/.uncoded/namespace.yaml
@@ -285,6 +285,7 @@ tests/:
       test_rendered_stub_has_no_line_range_comments:
       test_async_function_prefix:
       test_render_param_covers_input_kind:
+      test_return_annotation_rendered:
       test_class_with_bases:
       test_class_with_no_members_renders_body:
       test_attribute_with_annotation:
--- a/.uncoded/stubs/tests/test_stubs.pyi
+++ b/.uncoded/stubs/tests/test_stubs.pyi
@@ -89,7 +89,7 @@ class TestRenderStub:
     def test_async_function_prefix(self):
         ...
 
-    def test_function_with_annotations(self):
+    def test_render_param_covers_input_kind(self, params, expected):
         ...
 
     def test_class_with_bases(self):
--- a/.uncoded/stubs/tests/test_stubs.pyi
+++ b/.uncoded/stubs/tests/test_stubs.pyi
@@ -92,6 +92,9 @@ class TestRenderStub:
     def test_render_param_covers_input_kind(self, params, expected):
         ...
 
+    def test_return_annotation_rendered(self):
+        ...
+
     def test_class_with_bases(self):
         ...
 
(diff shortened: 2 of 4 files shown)


alimanfoo:
**Recommendation:** looks good, no findings.

The contract claimed by GH118 — every `_render_param` input shape defended by both compile smoke and a focused pin — is satisfied. The eight parametrized rows in `test_render_param_covers_input_kind` exhaust the three branches of `_render_param` (`/`-or-`*`, annotated, bare) across the shapes `_extract_params` produces (positional, varargs, kwargs, both separators), and the smoke source now adds `def collect(*args, **kwargs)` and `def separators(x, /, *, y)` so compile-level coverage of bare varargs/kwargs and both separators is in place too. Splitting the return-annotation pin out into its own `test_return_annotation_rendered` cleanly resolves the double-duty Grace flagged.

The GH119 drop is sound: `StubAssignment.value_source` defaults to `None`, so the kwarg in `_property_attribute` was asserting a default the dataclass already guarantees, the same shape as the `is_type_alias=False` drop in GH117.

PR #124, opened 2026-05-13
[claude] tighten _extract_params test coverage (#122)

### LINKED ISSUE #122: [claude] test_kwargs_and_varargs would not catch _extract_params drift
A regression in `_extract_params` that produces extra, reordered, or duplicated `StubParam`s in the varargs/kwargs case would slip past `tests/test_stubs.py::TestExtractStub::test_kwargs_and_varargs`. The assertions are membership-only:

```
assert StubParam("*args", "str") in f.params
assert StubParam("**kwargs", "int") in f.params
```

Every other collection assertion in `TestExtractStub` uses equality on the full collection — `test_function_no_annotations` (`f.params == [...]`), `test_class_with_attributes_and_methods` (`[(a.name, a.annotation) for a in cls.attributes] == [...]`), `test_imports_collected` (`module.imports == [...]`), `test_class_with_bases`, `test_tuple_unpacking_skipped`, `test_property_setter_and_deleter_suppressed`. The varargs/kwargs test is the only outlier.

Suggested direction: tighten to equality on the full params list, matching the shape of the rest of `TestExtractStub`.

Related to #113 and #118 — the same family of test-contract laxity (tests that pass regressions they should catch), different mechanism, different test class. The just-merged contract-naming for `_render_param` rendering did not survey `TestExtractStub` for the sibling pattern on the extraction side.

alimanfoo:
Closes #122.

`tests/test_stubs.py::TestExtractStub::test_kwargs_and_varargs` was the only collection assertion in `TestExtractStub` using membership (`in`) — a regression in `_extract_params` that produced extra, reordered, or duplicated `StubParam`s in the varargs/kwargs case would have slipped past. Tightened to full-list equality, matching the rest of the class.

While there, the recurrence pattern across #113, #118, and #122 (test-contract laxity in `test_stubs.py`) pointed at an unnamed contract on the extraction side. #118 named it on the rendering side — every `_render_param` branch is defended by a focused pin via `test_render_param_covers_input_kind`. This PR adds the parallel on the extraction side: `test_extract_params_covers_input_kind` walks every branch of `_extract_params` (regular positional, varargs, kwargs, posonly `/`, keyword-only `*`, kwonly) with one parametrize case each, asserting full-list equality on the extracted params. Future branches added to `_extract_params` are now expected to come with their matching case.

Two smaller tidies travel with it: the new test's parametrize IDs are aligned with the rendering-side siblings (`varargs_*`, `kwargs_*`), and the methods half of `test_class_with_attributes_and_methods` is tightened to projection equality to match the attributes half within the same test.

### DIFF
--- a/.uncoded/namespace.yaml
+++ b/.uncoded/namespace.yaml
@@ -264,6 +264,7 @@ tests/:
       test_class_with_bases:
       test_class_no_bases:
       test_kwargs_and_varargs:
+      test_extract_params_covers_input_kind:
       test_imports_collected:
       test_syntax_error_raises:
       test_source_order_preserved:
--- a/.uncoded/stubs/tests/test_stubs.pyi
+++ b/.uncoded/stubs/tests/test_stubs.pyi
@@ -31,6 +31,9 @@ class TestExtractStub:
     def test_kwargs_and_varargs(self):
         ...
 
+    def test_extract_params_covers_input_kind(self, source, expected):
+        ...
+
     def test_imports_collected(self):
         ...
 
--- a/tests/test_stubs.py
+++ b/tests/test_stubs.py
@@ -117,6 +117,66 @@ class TestExtractStub:
             StubParam("**kwargs", "int"),
         ]
 
+    @pytest.mark.parametrize(
+        "source,expected",
+        [
+            pytest.param(
+                "def f(x): pass\n",
+                [StubParam("x")],
+                id="regular_positional_bare",
+            ),
+            pytest.param(
+                "def f(x: str): pass\n",
+                [StubParam("x", "str")],
+                id="regular_positional_annotated",
+            ),
+            pytest.param(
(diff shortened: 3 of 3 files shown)


alimanfoo:
**Recommendation:** approve subject to one nit.

## Nits
1. `tests/test_stubs.py:154,158,162,166` — the new IDs `posonly_bare/annotated` and `kwonly_bare/annotated` abbreviate, while the render-side siblings in `test_render_param_covers_input_kind` spell out `positional_only_separator` and `keyword_only_separator`. The varargs/kwargs IDs already match across both sides. Aligning these (`positional_only_bare/annotated`, `keyword_only_bare/annotated`) would make the parallel exact and let a reader pivot between the two tests on shared identifiers. Minor — the render-side cases test the separator alone, while the extract-side cases test arg+separator together, so the wording isn't a one-to-one mapping and the abbreviation is defensible.

PR #127, opened 2026-05-13
[claude] tighten _extract_params test coverage (#125)

### LINKED ISSUE #125: [claude] branch combinations and contract claims slip past _extract_params tests
The recurrence on `tests/test_stubs.py` and the `_extract_params` / `_render_param` surface — #113, #115, #118, #122 — has been addressed branch-by-branch each time. Each pass closed the specific shape it named (substring → compile-smoke, branch-by-branch parametrize on rendering, then on extraction). The contract that "every code path through `_extract_params` is exercised" is still implicit. Two specific gaps surfaced in the post-merge sweep on #122 (PR #124):

**Branch combination not pinned: vararg + kwonly.** `_extract_params` (`src/uncoded/stubs.py` L69–101) suppresses the synthetic `*` separator when a vararg is present (the `elif args.kwonlyargs` branch). The new `test_extract_params_covers_input_kind` covers vararg-alone and kwonly-alone cases individually, but not the combination `def f(*args, x): pass` where the vararg branch fires and the kwonly args are appended without the `*` separator. A regression that wrongly emitted `*` after a vararg would not be caught.

**Positive contract claim not pinned: defaults are dropped.** `_extract_params`'s docstring says "without defaults" and the implementation never touches `args.defaults` or `args.kw_defaults`. No test pins this. A regression that started encoding defaults into `StubParam.name` or `StubParam.annotation` (for example `StubParam("x=0", "int")`) would render as invalid Python and is not caught by any current test, including the compile-smoke in `test_renders_valid_python_for_representative_source` whose representative source has no functions with defaults.

Suggested direction: extend `test_extract_params_covers_input_kind` with a parametrize case for the vararg+kwonly combination (`def f(*args, x): pass` → `[StubParam("*args"), StubParam("x")]`) and a parametrize case for defaults (`def f(x: int = 0): pass` → `[StubParam("x", "int")]`). A complementary direction is to add a function with defaults to the representative source so the compile-smoke covers it too.

alimanfoo:
Closes #125.

Four new parametrize cases in `test_extract_params_covers_input_kind` pin two coverage gaps:

- `vararg_with_kwonly` (`def f(*args, x): pass`) pins the branch interaction where the vararg branch fires *and* the kwonly loop runs *without* the `*` separator. Existing single-feature cases didn't cover this combination, so a regression wrongly emitting the separator after a vararg or dropping the kwonly args after a vararg would not have been caught.
- `defaults_dropped_positional` and `defaults_dropped_kwonly` pin the "defaults dropped" claim in `_extract_params`'s docstring. The implementation reads neither `args.defaults` nor `args.kw_defaults`; before this change no test caught a regression encoding a default into name or annotation. Compile-smoke does not catch every form because `def f(x: int = 0):` is valid Python.

The legacy `test_kwargs_and_varargs` is deleted. A new bare `vararg_with_kwarg` case (`def f(*args, **kwargs): pass`) takes its place in the parametrize table; annotated-combination coverage is retained via compile-smoke on `def build(*args: str, **kwargs: int)` in the representative source.

The representative source in `test_renders_valid_python_for_representative_source` gains a function with a default (`def configure(retries: int = 3) -> None: pass`), extending compile-smoke coverage of the defaults contract.

### DIFF
--- a/.uncoded/namespace.yaml
+++ b/.uncoded/namespace.yaml
@@ -263,7 +263,6 @@ tests/:
       test_class_with_attributes_and_methods:
       test_class_with_bases:
       test_class_no_bases:
-      test_kwargs_and_varargs:
       test_extract_params_covers_input_kind:
       test_imports_collected:
       test_syntax_error_raises:
--- a/.uncoded/stubs/tests/test_stubs.pyi
+++ b/.uncoded/stubs/tests/test_stubs.pyi
@@ -28,9 +28,6 @@ class TestExtractStub:
     def test_class_no_bases(self):
         ...
 
-    def test_kwargs_and_varargs(self):
-        ...
-
     def test_extract_params_covers_input_kind(self, source, expected):
         ...
 
--- a/tests/test_stubs.py
+++ b/tests/test_stubs.py
@@ -103,18 +103,6 @@ class TestExtractStub:
         module = extract_stub(source, "pkg/plain.py")
         assert module.classes[0].bases == []
 
-    def test_kwargs_and_varargs(self):
-        source = textwrap.dedent("""\
-            def build(*args: str, **kwargs: int) -> None:
-                pass
-        """)
-        module = extract_stub(source, "pkg/build.py")
-        f = module.functions[0]
-        assert f.params == [
-            StubParam("*args", "str"),
-            StubParam("**kwargs", "int"),
-        ]
-
     @pytest.mark.parametrize(
         "source,expected",
         [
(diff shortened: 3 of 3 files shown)


alimanfoo:
**Recommendation:** looks good, one small accuracy nit.

## Nits
1. The PR description says `vararg_with_kwarg` "consolidates" the legacy `test_kwargs_and_varargs` and that "its content lives in the parametrize table." The legacy case was annotated (`*args: str, **kwargs: int`); the new parametrize case is bare (`*args, **kwargs`). The annotated combination survives only via compile-smoke on the `build` function in the representative source, not as a unit-test case. The shape change is fine — annotated `*args` and annotated `**kwargs` are each pinned separately, so the per-branch annotation handling is covered — but "consolidates" slightly overstates the continuity. A more accurate phrasing would say the legacy case is replaced by a bare combination case, with annotated-combination coverage retained via compile-smoke.

PR #128, opened 2026-05-14
[claude] add test coverage check to the quality bar (#126)

### LINKED ISSUE #126: [claude] quality bar does not measure test coverage
uncoded's quality bar runs lint, format, type-check, tests, and index sync via pre-commit hooks. None of these measure test coverage. Uncovered branches and dead code paths surface only when someone reads the code and notices the gap — for example, #118 and #125 both name specific coverage gaps that a coverage report would have flagged automatically (untouched branches, code paths no test exercises).

A coverage check would not catch every kind of gap. Branch combinations and positive contract claims (the #125 cases) are not measured by line/branch coverage tooling. But it would surface a different class of regression: lines or branches with zero test coverage, dead code creeping in, and gradual drift as new code lands without tests.

Suggested direction: add `pytest-cov` (or `coverage.py`) to dev dependencies and run `pytest --cov=src --cov-branch --cov-report=term-missing --cov-fail-under=<N>` as part of the quality bar (pre-commit and CI). Pick an initial threshold around current coverage and let it ratchet up over time. Consider whether to include `tests/` in the measured source so helper code in tests is also covered.

alimanfoo:
Adds a branch coverage gate to the quality bar. Pytest now reports coverage
on every run and fails when coverage drops below the configured threshold.
The gate runs in CI without any workflow change — pytest picks up the
config from `pyproject.toml`.

The initial threshold is `fail_under = 97` (current measured coverage
97.56%). It can ratchet up over time as uncovered branches close.

To run a subset of tests without the coverage gate locally, pass
`--no-cov`:

```
uv run pytest tests/test_stubs.py --no-cov
```

AGENTS.md and README.md both name the gate alongside the existing
`uv run pytest` reference so future contributors aren't surprised by it.

Closes #126


alimanfoo:

## Nits
1. `README.md:213` and `AGENTS.md:50` fold the new context into a parenthetical that chains two clauses with a semicolon ("branch coverage is enforced; see `[tool.coverage.report]` in `pyproject.toml` for the threshold"). The information is correct, but the line is dense to skim. A short follow-up sentence after the code block — e.g. "The threshold lives in `[tool.coverage.report]` of `pyproject.toml`." — would lower the density without losing the pointer.

2. The `--no-cov` example pins on `tests/test_stubs.py` in both `README.md:223` and `AGENTS.md:54`. The example is concrete, which is good, but it couples the doc to one specific test file. A rename of that file later would leave stale text. Minor — keeping a real path is reasonable.

## Out of scope but noticed
1. `README.md:225` still labels `uv run pre-commit run --all-files` as "all checks (the same suite CI runs)", but CI also runs `pytest -v` (`.github/workflows/ci.yml:33`), now with the coverage gate behind it. The line pre-dates this PR, but the gap between that claim and what CI actually runs is wider now that tests carry a real gate. A future tidy could list both commands or rephrase.

PR #130, opened 2026-05-14
[claude] close residual src/ coverage gaps and ratchet gate to 100 (#129)

### LINKED ISSUE #129: [claude] residual coverage gaps in src/ surfaced by the new coverage gate
The new branch coverage gate landed at `fail_under = 97` (#126; measured baseline 97.56%). The gate defends against drops from here, but ~2.5% of branches and lines in `src/` are still uncovered today — they sit beneath the threshold. Closing them strengthens the gate and lets the threshold ratchet up.

The `term-missing` report from `uv run pytest` flags the gaps. Initial sighting:

- `src/uncoded/stubs.py` (96% file coverage): lines 123, 132, 315, 364-366; partial branches 182->179, 216->209, 376->372.
- `src/uncoded/cli.py`: lines 111-112.
- `src/uncoded/extract.py`: line 51; partial branches 40->37, 71->68.
- `src/uncoded/serena_setup.py`: line 184.

Suggested direction: investigate each gap and either add a test that pins the branch/line, or remove the path if it's unreachable. As gaps close, ratchet `fail_under` up.

alimanfoo:
The 97% coverage gate added in #126 defends against future drops but left ~2.5% of `src/` uncovered at the time it landed. #129 listed those residual gaps. This PR closes each one and ratchets `fail_under` to 100, so any future uncovered line or branch breaks the build.

Each gap was investigated and given one of three treatments:

- **Reachable, missing test → add a focused test** that exercises the branch or line. Tests live alongside their existing peers in `tests/test_stubs.py`, `tests/test_cli.py`, `tests/test_extract.py`, and `tests/test_serena_setup.py`.
- **Statically dead → remove the guard.** Two identical `isinstance(node.name, ast.Name)` guards on `ast.TypeAlias` — one in `src/uncoded/stubs.py`, one in `src/uncoded/extract.py`. The CPython AST grammar types `TypeAlias.name` as `Name`, so the guards could never fire.
- **Race-condition path → `# pragma: no cover`.** The orphan-cleanup loop in `_write_stubs` counts a removal only when `remove_file` returns `True`. The `False` arm is only reached if a `.pyi` is deleted between `rglob` discovering it and `remove_file` trying to unlink it — not exercisable without monkey-patching, and the existing `if`-arm covers the contract.

After closing the gaps, `fail_under` in `pyproject.toml` moves from `97` to `100`.

Closes #129.


alimanfoo:
**Recommendation:** approve subject to one nit.

## Nits
1. `src/uncoded/stubs.py` lines 376–377: the `else: pass  # pragma: no cover` exists only to host the pragma — the `else` body is a no-op. Coverage's `# pragma: no branch` on the `if` at line 374 is the idiomatic way to say "the False arm isn't measurable, the True arm is fine," and it avoids the dead `else`. The race-condition comment can move inline or onto a comment line above the `if`.

alimanfoo (commented):
LGTM

Issue #111, opened 2026-05-11
[claude] proposal: replace Serena/MCP with uncoded body + refs CLIs (cross-platform)


alimanfoo:
## Summary

Replace Serena (MCP wrapping ty via solidlsp) with two small uncoded-owned CLIs — `uncoded body` for precise symbol-body extraction, and `uncoded refs` for impact analysis via a short-lived ty subprocess. Drop the entire MCP layer. Keep the existing static index (namespace + stubs) unchanged. Net result: smaller install, no MCP layer, no third-party LSP middleware, fully cross-platform (Claude, Codex, any agent with Read + shell-exec), and the staleness class of bugs that started this thread (#46) goes away by construction.

## Background

#46 reported `find_referencing_symbols` returning incomplete results after a structural move. Investigation traced the root cause to solidlsp's `LSPFileBuffer` keeping a sticky `_is_open_in_ls=True` flag after the first `didOpen`. External file changes (Edit/Write, formatter, manual edits) do not trigger `didChange` to ty, so ty's view of the document goes stale and reference queries return the old answer.

The upstream situation is partial: oraios/serena PR #1028 (merged) fixes the related `replace_content` stale-buffer corruption but does not propagate fresh content to the LSP. Serena issue #69 ("Monitor file system for external changes") was closed by the maintainer with "the language servers already handle this" — true for pyright/clangd/rust-analyzer (which run their own watchers), not for ty (which trusts the client per LSP spec).

## Drivers and the tool that serves each

uncoded was always built around three drivers:

| Driver | Need | Tool |
|---|---|---|
| 1. Recall (no guessing) | All symbols loaded in context at session start | uncoded `namespace.yaml` (existing) |
| 2a. API survey | Compact per-file API view | uncoded `.pyi` stubs (existing) |
| 2b. Body read | Precise byte-range extraction of one symbol | `uncoded body` CLI (new) |
| 3. Impact analysis | Semantic reference search | `uncoded refs` CLI (new) |

Drivers 1 and 2 are about minimising context waste; driver 3 is about completeness of analysis. Today driver 2b and driver 3 are served by Serena (`find_symbol(include_body=True)` and `find_referencing_symbols`). After this change, all three drivers are served by uncoded-owned CLIs over a static index — no MCP layer, no LSP middleware.

## Why CLI-only and not Claude Code's native LSP tool

[...]

PR #132, opened 2026-05-14
[claude] add uncoded body subcommand (#111)

alimanfoo:
## Summary

First step of #111's migration plan to replace Serena/MCP with native uncoded CLIs. Adds an `uncoded body` subcommand that prints the source body of a named Python symbol — for an agent that has read the stub signature and now needs the implementation.

The agent invokes:

```
uncoded body <name_path> --in <relative_path>
```

`name_path` follows the Serena convention: one segment for a top-level symbol (`fn`), two for a class member (`Class/method`). `--in` is required — the agent already has the file path from the stub it just read, and requiring it avoids an ambiguity-recovery path that has no real consumer.

Stdout is byte-identical to the source span on disk, so the bytes can be pasted straight into an Edit tool's \`old_string\`. Stderr is reserved for errors. The slicing is line-based — no \`ast.unparse\` normalisation — and decorators are included in the span so a method's full editable surface comes through as one unit.

Serena stays in place. This PR adds a capability; it doesn't remove anything. The companion \`refs\` subcommand and the Serena retirement come in later steps of #111.


alimanfoo:
**Recommendation:** approve subject to a few small things.

## Non-blocking
1. `_body` in `src/uncoded/cli.py` (lines 140–159) has no docstring, while its sibling `_sync` carries a full one describing return codes and behaviour. The contract for `_body` — what it writes to stdout, what triggers each `return 1` — is worth a short docstring for the same reason `_sync`'s is.

## Nits
1. The error-message path for `_body` (`src/uncoded/cli.py:148–155`) is inconsistent: the `BodyNotFound` branch prints `resolve_body`'s message, which embeds the project-resolved absolute path (`'foo' not found in /abs/project/src/foo.py`), while the `FileNotFoundError` and `SyntaxError` branches use the user-supplied relative `in_path` (`src/foo.py: file not found.`). Aligning on one form would read more uniformly to the agent that consumes stderr.
2. The top-level match condition in `resolve_body` (`src/uncoded/body.py:32–40`) mixes `and` and `or` in one expression and relies on Python's precedence to bind the ClassDef branch correctly. It's right, but it reads as a single tall expression where the two cases are doing quite different things. Splitting into two `if` clauses, or naming the cases (`matches_class = ...`, `matches_function = ...`), would be easier to scan.
3. The `main` docstring (`src/uncoded/cli.py:163–177`) packs all four subcommands into a single semicolon-chained sentence. Now that there are four, breaking it into one short sentence per subcommand, or a small bullet list, would recover the contract on first read.

src/uncoded/body.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+def resolve_body(name_path: str, in_path: Path) -> str:
+    """Return the source text for the symbol named by name_path in in_path.
+
+    name_path follows Serena convention: one segment for a top-level symbol,
+    two segments (Class/member) for a class method, property, or attribute.
+    Raises BodyNotFound if the symbol is not present. Lets FileNotFoundError
+    propagate if in_path does not exist, and SyntaxError if in_path cannot
+    be parsed.
```
alimanfoo: We will ultimately be getting rid of Serena, so referring to Serena here in the docstring will go stale. Write for a reader who has never heard of Serena.

README.md
```diff
@@ -114,6 +114,27 @@ file is byte-identical to what a rebuild would produce, and 1 otherwise
...
+
+```
+uncoded body <name_path> --in <relative_path>
+```
+
+`name_path` follows Serena convention: one segment (`fn`) for a top-level
+symbol, two for a class member (`Class/method`). `--in` is the source file's
+path relative to the project root. The command prints the source text of the
+symbol to stdout, byte-identical to what's on disk — no reformatting, no
+`ast.unparse` normalisation.
```
alimanfoo: Again, don't mention Serena here.

PR #137, opened 2026-05-15
[claude] uncoded body: cwd-relative --in and validated name_path shapes (#134, #135)

### LINKED ISSUE #134: [claude] uncoded body silently mishandles unsupported name_path shapes
`uncoded body <name_path>` accepts any string and runs the AST walk, but only one-segment (`foo`) and two-segment (`Class/member`) paths can actually match. Other shapes fall through to `BodyNotFound`, which tells the user "symbol not found" when the real issue is that the path syntax isn't supported.

`src/uncoded/body.py:resolve_body` and `_resolve_class_member`: three-segment paths (`A/B/C`) are split into `head="A"` and `tail="B/C"`, and `tail` is then compared verbatim against class-member names — never matching. Nested classes (e.g. `Outer/Inner/method`) hit the same fallthrough because `_resolve_class_member` doesn't recurse into `ClassDef` children.

Direction: define the supported shapes explicitly (likely the same set the namespace map enumerates — top-level symbols and one level of class membership). Reject any other shape with an error that names the supported syntax. If nested classes should become supported, that expands both the resolver and the index together — but that's a separate decision.

### LINKED ISSUE #135: [claude] uncoded body's --in path is silently anchored at the project root
An agent running `uncoded body fn --in foo.py` expects the `--in` path to be relative to wherever they invoked the command. The actual behaviour is project-root-anchored — `<project_root>/foo.py`, regardless of cwd. If a different file with the same basename happens to exist at the project root, the agent gets `BodyNotFound` ("symbol not in this file") when the real failure is "wrong file." The README documents the contract; the behaviour itself is silent.

`src/uncoded/cli.py:_body` resolves `target = project_root / in_path` after `_find_project_root` walks up to `pyproject.toml`. The cwd plays no role in the path resolution.

Direction: either change the contract so `--in` resolves against cwd (matches Unix tool conventions, removes the silent failure mode), or echo the anchored path so the user sees what was actually opened. Other cures are possible; the concern is acknowledging the choice rather than mandating a specific one.

alimanfoo:
## Summary

Two silent failures in `uncoded body`:

- `--in PATH` resolved against the project root rather than cwd, so callers running from a subdirectory got the wrong file opened. If a file with the matching basename happened to exist at the project root, they got the wrong body returned with no indication anything was off.
- `name_path` shapes with more than two segments or empty segments (e.g. `A/B/C`, `Outer/Inner/method`, `/foo`) fell through to a generic "symbol not found" error, even though the contract only supports `name` and `Class/member`.

This PR:

- Makes `--in` resolve against cwd (Unix unmarked default). `_body` no longer needs `_find_project_root`, and the "no pyproject.toml" return path goes away — the command is project-agnostic now.
- Adds `UnsupportedNamePath`, raised at the entry of `resolve_body` for any input that doesn't have exactly one or two non-empty segments. The CLI handler prints the exception's message, which names the supported shapes.

Closes #134
Closes #135


alimanfoo:

## Non-blocking

1. `src/uncoded/cli.py:231` — the `--in` help text was shortened to just "Source file path." with no anchor. Before, it told the user the path was project-root-relative; now it says nothing about how the path is resolved. Since this PR deliberately changes that contract (and the README was updated to say "relative to cwd"), the CLI help text should match, e.g. "Source file path (relative to current directory)." A user running `uncoded body --help` is the most likely reader of this line.

## Nits

1. `src/uncoded/body.py:32-40` — `name_path.split("/")` runs twice in `resolve_body`: once for shape validation, once again with `maxsplit=1` to recover head/tail. The validation already produced a fully-split list, so deriving `head` and `tail` from `all_segments` directly would avoid the second split and remove a small inconsistency (the validator sees the full segment list, the matcher sees only the first split).

2. `tests/test_body.py:288` — `pytest.raises(UnsupportedNamePath, match="'name'")` and the subsequent explicit `for shape in SUPPORTED_SHAPES` loop check overlapping things. Either form alone covers the intent; keeping both is a bit of belt-and-braces noise.

3. `tests/test_cli.py:477` — the renamed `test_in_path_resolves_relative_to_cwd` still calls `_init_repo`, which creates a `pyproject.toml`. Since the headline behaviour change is that `_body` is now project-agnostic, a test that demonstrates it working *without* any pyproject.toml in scope would be a stronger anchor on that property. Optional.

alimanfoo:
All four findings accepted and addressed in 2ffe470.

## Non-blocking

1. **`--in` help text** — accepted. Help now reads `"Source file path (relative to current directory)."` to match the README contract.

## Nits

1. **Double split in `resolve_body`** — accepted. `head` and `tail` are now derived from the `all_segments` list produced by the validation block; the redundant `split("/", maxsplit=1)` is gone.
2. **Belt-and-braces check in `_assert_raises`** — accepted. Dropped the `match=` argument; the explicit loop over `SUPPORTED_SHAPES` remains, since it checks both supported-shape strings rather than just one.
3. **Project-agnostic body test** — accepted. Added `test_works_without_project_root` in `TestBodyCommand` — bare `tmp_path`, no `pyproject.toml`, asserts `_body` returns 0 and the correct body.

PR #138, opened 2026-05-15
[claude] cut over body retrieval to uncoded body (#111)

alimanfoo:
## Summary

Agents that work in this repo have been routing body reads through Serena's
`find_symbol(include_body=True)`. This PR moves those routes onto the existing
`uncoded body` CLI, so the tool gets real-world use during the rest of #111's
migration. Scope is the body-retrieval surface only — reference search, edit,
rename, and safe-delete still go through Serena.

## What changes

- **Instruction prose in `CLAUDE.md`/`AGENTS.md`** (the injected
  `<!-- uncoded:start -->` section). The dispatch-rule example, Step 3
  heading, body-retrieval bullet, and related cross-references now point at
  `uncoded body`. Serena keeps the four operations it still serves (find
  references, rename, edit, safe-delete).

- **Coherence-review skill** (`.claude/skills/coherence-review/SKILL.md`,
  `.agents/skills/coherence-review/SKILL.md`). Six body-read sites flipped.
  The "When confidence needs the body" sub-section flattens to one step
  because `uncoded body` returns the whole body in a single call (Serena's
  call returned the docstring and made the body available as a second read).

- **`README.md`.** The "How agents use it" walkthrough, the coherence-review
  description, and the language-server-operations list all updated.

- **Serena allowlist.** `mcp__serena__find_symbol` removed from
  `SERENA_ALLOWED_TOOLS` since no documented workflow uses it post-cutover.
  `_sync_claude_settings` now prunes stale `mcp__serena__*` entries from
  `.claude/settings.json` so the source list and the generated file stay in
  sync in both directions (previously the function only appended; the gap
  was invisible to the dogfooding test, which only checked one direction).
  The dogfooding test was tightened to check both directions on the Serena
  prefix.

## Relation to #111

#111 lays out a five-step migration to replace the Serena/MCP layer with two
small uncoded-owned CLIs. This is step 3 in that plan, reordered to land
before step 2 (`uncoded refs`) so the body CLI gets real usage from agents
during the rest of the work. #111 stays open as the umbrella tracker until
refs lands and the full Serena setup can be retired.


alimanfoo:
**Recommendation:** approve subject to nits — two prose loose ends in the dispatch-rule section should be tightened before merge.

## Non-blocking

1. The closing summary line of the dispatch-rule section still routes symbol names to Serena. In `src/uncoded/instruction_files.py:137-138` (mirrored into `CLAUDE.md` and `AGENTS.md`):

   > The dispatch rule turns on the search term: a symbol name → Serena; a regex or free-text phrase → grep.

   After this PR, symbol-name body reads route to `uncoded body`, not Serena. The line is the headline restatement of the rule and now contradicts the Step 3 bullets that precede it. This is the same edit the PR is making, missed on the closing sentence. Something like "a symbol name → the index" or "a symbol name → `uncoded body` or Serena" would restore agreement with the bullets.

2. The Step 3 heading uses "find" without saying *what* is found. In `src/uncoded/instruction_files.py:80-81`:

   > Step 3 — Act. Use `uncoded body` to read a symbol's body; use Serena to find, rename, edit, and delete symbols.

   Before this PR, "find" mapped to Serena's `find_symbol`. The PR drops `find_symbol` from the allowlist, so the only Serena tool with "find" in its name is now `find_referencing_symbols` (find callers). A reader who reads the heading and stops there could plausibly infer `find_symbol` is still in play. Spelling it out — "find callers" or "find references" — would tie the heading to the bullets that follow.

## Nits

1. "each Serena tool needs" no longer covers all the bullets. In `src/uncoded/instruction_files.py:82-84`:

   > you have the exact `relative_path` and `name_path` each Serena tool needs (`ClassName/method` for a method, `function_name` for a top-level function)

   The first bullet under Step 3 is now `uncoded body`, which also takes `name_path` and `--in <relative_path>` in the same shape. "each tool" would cover the Serena tools and `uncoded body` consistently.

alimanfoo:
All three findings accepted and addressed in a204a0d.

## Non-blocking

1. **Closing summary of the dispatch rule.** Updated to "a symbol name → the index; a regex or free-text phrase → grep". "The index" now covers the namespace map, stubs, `uncoded body`, and Serena, in agreement with the Step 3 bullets above.

2. **Step 3 heading.** Updated to "use Serena to find callers, rename, edit, and delete symbols" — names what the remaining Serena tool actually finds (`find_referencing_symbols`), tied to the bullet below.

## Nits

1. **"each Serena tool needs" generalization.** Updated to "each tool needs" — covers Serena and `uncoded body` consistently, since both take the same `name_path` / `--in <relative_path>` coordinates.

Issue #140, opened 2026-05-15
[claude] find_symbol left in Serena's exposed tool surface via excluded_tools after allowlist removal


alimanfoo:
#138 removed `mcp__serena__find_symbol` from `SERENA_ALLOWED_TOOLS` and added prune behaviour to `_sync_claude_settings` so `.claude/settings.json` stays in sync with the source list. But the project YAML's `excluded_tools` layer wasn't updated — `find_symbol` is still in Serena's exposed tool surface for any non-Claude MCP client.

The module docstring in `src/uncoded/serena_setup.py` describes a defense-in-depth architecture:

> The exclusions in `SERENA_PROJECT_FIELDS` defend in depth alongside the prose ... in `instruction_files._SECTION_BODY`: the prose tells the agent not to call those tools, and the project config drops them from Serena's exposed surface so the call cannot land even if the prose is missed.

After the cutover, `find_symbol` is in the "no documented workflow uses it" set but only blocked at the allowlist layer:

- `instruction_files._SECTION_BODY` no longer routes agents to `find_symbol` ✓
- `SERENA_ALLOWED_TOOLS` removed it ✓
- `SERENA_PROJECT_FIELDS["excluded_tools"]` still doesn't list it ✗

This has a behaviour consequence: the allowlist is Claude Code-specific (`.claude/settings.json`). `.serena/project.yml` controls what Serena exposes to *any* MCP client. For Codex or any other MCP client without the allowlist gate, Serena still exposes `find_symbol` — the cutover only landed for Claude Code.

Adding `"find_symbol"` to `SERENA_PROJECT_FIELDS["excluded_tools"]` in `src/uncoded/serena_setup.py` (and re-running `uncoded setup` to regenerate `.serena/project.yml`) would complete the defense-in-depth narrowing for `find_symbol`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #46, opened 2026-04-30
[claude] find_referencing_symbols returned incomplete results after structural move


alimanfoo:
During a session task that moved `read_project_name` from `src/uncoded/serena_setup.py` to `src/uncoded/config.py` (commit `3883cfc` on PR #45), `mcp__serena__find_referencing_symbols` queried against the moved function returned exactly **one** reference — the call site inside `setup()` in `src/uncoded/serena_setup.py`. It did not return the four references in `tests/test_config.py`:

- The import in `from uncoded.config import (..., read_project_name, ...)`
- Three call sites inside `TestReadProjectName::test_*`

A repo-wide grep found those four immediately.

This contradicts what `CLAUDE.md` claims of `find_referencing_symbols`:

> Returns every reference resolved by the language server — every import, every call site, every re-export. Grep on the name misses re-exports and adds false positives from comments, strings, and attribute lookups on other types. If the next move depends on the answer being complete, grep cannot give you that.

The Serena-vs-grep dispatch rule depends on the completeness of `find_referencing_symbols`. If there is a known staleness window after structural moves, an LSP limitation under specific conditions, or a real bug, agents need to know — either to re-query after a beat or to fall back to a grep cross-check after structural moves.

Possible causes worth investigating:

- LSP indexing staleness immediately after a `git mv` / function move
- An interaction between source-file location and test-file location
- A specific Serena / ty / LSP version issue

Reproducer: the chain of `find_symbol` + `rename_symbol` + `insert_after_symbol` operations on `src/uncoded/serena_setup.py` and `src/uncoded/config.py` from commit `3883cfc`, followed by a `find_referencing_symbols` query on `read_project_name`.

Surfaced by the maintainer agent during a coherence review.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

alimanfoo:
Reinforcing this with a related observation from PR #108 task 4.

Working on PR #108 task 4 (drop `generate_section`), the workflow was: edit seven test call sites in `tests/test_instruction_files.py` via the Edit tool to reference `SECTION` directly, then call `safe_delete_symbol generate_section` to remove the now-orphan function.

[...]

PR #145, opened 2026-05-16
[claude] add uncoded refs, retire find_referencing_symbols (#111)

alimanfoo:
Step 2 of #111's migration from Serena/MCP to native uncoded CLIs, plus the matching dispatch-rule cutover. After step 1 added `uncoded body` (#132), this step adds `uncoded refs` for impact analysis — find every reference to a named symbol — and removes `find_referencing_symbols` from Serena's exposed surface in this repo.

Each `uncoded refs` invocation spawns a fresh `ty server` over stdio, opens the file from disk, queries `textDocument/references`, and exits. No long-lived LSP buffer to go stale — the class of bug behind #46 doesn't exist on this path.

The agent dispatch rule across `_SECTION_BODY`, `README.md`, `_SKILL_CONTENT`, and AGENTS.md's hand-written preamble now routes the "find callers" step to `uncoded refs`. Serena's role narrows to renaming, editing by symbol, and safe-deletion. Serena's surface is closed in two layers: `find_referencing_symbols` is removed from `SERENA_ALLOWED_TOOLS` and added to `excluded_tools` in `SERENA_PROJECT_FIELDS` — the same defence-in-depth shape `find_symbol` follows after #140.

Internally the AST walker that maps `name_path` to `ast.stmt` was lifted from `body.py` into a shared helper so `body` and `refs` share it. Source is read once per public-function call.

`ty` pinned at 0.0.37. The remaining Serena tools (rename, safe-delete, edit-by-symbol) plus `serena_setup.py` and `SERENA_VERSION` are out of scope here — they belong to step 5 of #111.

### DIFF
--- a/src/uncoded/body.py
+++ b/src/uncoded/body.py
@@ -29,12 +29,41 @@ def resolve_ast_node(name_path: str, in_path: Path) -> ast.stmt:
     FileNotFoundError propagate if in_path does not exist, and SyntaxError if
     in_path cannot be parsed.
     """
+    source = in_path.read_text()
+    return _resolve_ast_node_from_source(
+        name_path=name_path, source=source, in_path=in_path
+    )
+
+
+def resolve_body(name_path: str, in_path: Path) -> str:
+    """Return the source text for the symbol named by name_path in in_path.
+
+    name_path is a slash-separated path: a single segment names a top-level
+    symbol; two segments name a class member as Class/member (a method,
+    property, or attribute).
+    Raises UnsupportedNamePath if name_path has more than two segments or any
+    empty segment. Raises BodyNotFound if the symbol is not present. Lets
+    FileNotFoundError propagate if in_path does not exist, and SyntaxError if
+    in_path cannot be parsed.
+    """
+    source = in_path.read_text()
+    node = _resolve_ast_node_from_source(
+        name_path=name_path, source=source, in_path=in_path
+    )
+    return _extract_body(node=node, lines=source.splitlines(keepends=True))
+
+
+def _resolve_ast_node_from_source(
+    *,
+    name_path: str,
+    source: str,
+    in_path: Path,
+) -> ast.stmt:
     all_segments = name_path.split("/")
     if len(all_segments) > 2 or any(s == "" for s in all_segments):
         raise UnsupportedNamePath(
             f"Unsupported name_path {name_path!r}: use 'name' or 'Class/member'"
         )
-    source = in_path.read_text()
     tree = ast.parse(source, filename=str(in_path))
     head = all_segments[0]
     tail = all_segments[1] if len(all_segments) == 2 else None
@@ -69,22 +98,6 @@ def resolve_ast_node(name_path: str, in_path: Path) -> ast.stmt:
         )
 
     return top_match
-
-
-def resolve_body(name_path: str, in_path: Path) -> str:
-    """Return the source text for the symbol named by name_path in in_path.
-
-    name_path is a slash-separated path: a single segment names a top-level
-    symbol; two segments name a class member as Class/member (a method,
-    property, or attribute).
-    Raises UnsupportedNamePath if name_path has more than two segments or any
-    empty segment. Raises BodyNotFound if the symbol is not present. Lets
-    FileNotFoundError propagate if in_path does not exist, and SyntaxError if
-    in_path cannot be parsed.
-    """
-    node = resolve_ast_node(name_path, in_path)
-    lines = in_path.read_text().splitlines(keepends=True)
-    return _extract_body(node=node, lines=lines)
 
 
 def _resolve_class_member(
--- a/src/uncoded/body.py
+++ b/src/uncoded/body.py
@@ -3,7 +3,7 @@
 import ast
 from pathlib import Path
 
-from uncoded.resolver import NamePath, _resolve_ast_node_from_source
+from uncoded.resolver import NamePath, resolve_ast_node_from_source
 
 
 def resolve_body(name_path: NamePath, in_path: Path) -> str:
@@ -14,7 +14,7 @@ def resolve_body(name_path: NamePath, in_path: Path) -> str:
     in_path cannot be parsed.
     """
     source = in_path.read_text()
-    node = _resolve_ast_node_from_source(
+    node = resolve_ast_node_from_source(
         name_path=name_path, source=source, in_path=in_path
     )
     return _extract_body(node=node, lines=source.splitlines(keepends=True))
--- a/src/uncoded/body.py
+++ b/src/uncoded/body.py
@@ -6,7 +6,7 @@ from pathlib import Path
 from uncoded.ast_helpers import assign_target_name, property_kind
 
 
-class BodyNotFound(Exception):
+class SymbolNotFound(Exception):
     """Raised when name_path cannot be found in the given file."""
 
 
@@ -25,7 +25,7 @@ def resolve_ast_node(name_path: str, in_path: Path) -> ast.stmt:
     symbol; two segments name a class member as Class/member (a method,
     property, or attribute).
     Raises UnsupportedNamePath if name_path has more than two segments or any
-    empty segment. Raises BodyNotFound if the symbol is not present. Lets
+    empty segment. Raises SymbolNotFound if the symbol is not present. Lets
     FileNotFoundError propagate if in_path does not exist, and SyntaxError if
     in_path cannot be parsed.
     """
@@ -41,7 +41,7 @@ def resolve_name_position(name_path: str, in_path: Path) -> tuple[int, int]:
     Follows LSP convention: both line and character are 0-indexed.
     For def/async def/class, character points past the keyword to the identifier.
     For assignments and type aliases, character points at the start of the target name.
-    Raises UnsupportedNamePath, BodyNotFound, FileNotFoundError, and SyntaxError
+    Raises UnsupportedNamePath, SymbolNotFound, FileNotFoundError, and SyntaxError
     under the same conditions as resolve_ast_node.
     """
     node = resolve_ast_node(name_path, in_path)
@@ -69,7 +69,7 @@ def resolve_body(name_path: str, in_path: Path) -> str:
     symbol; two segments name a class member as Class/member (a method,
     property, or attribute).
     Raises UnsupportedNamePath if name_path has more than two segments or any
-    empty segment. Raises BodyNotFound if the symbol is not present. Lets
+    empty segment. Raises SymbolNotFound if the symbol is not present. Lets
     FileNotFoundError propagate if in_path does not exist, and SyntaxError if
     in_path cannot be parsed.
     """
@@ -114,7 +114,7 @@ def _resolve_ast_node_from_source(
             top_match = node
 
     if top_match is None:
-        raise BodyNotFound(f"{name_path!r} not found in {in_path}")
+        raise SymbolNotFound(f"{name_path!r} not found in {in_path}")
 
     if tail is not None and isinstance(top_match, ast.ClassDef):
         return _resolve_class_member(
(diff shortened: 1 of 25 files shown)


alimanfoo:
**Recommendation:** approve subject to non-blocking items below; nothing blocks merge.

## Non-blocking

1. **Missing `uvx` is reported as "file not found" for the source file.** In `src/uncoded/cli.py:183`, `_refs` catches `FileNotFoundError` and prints `f"Error: {in_path}: file not found."`. But `subprocess.Popen(["uvx", ...])` in `query_references` also raises `FileNotFoundError` if `uvx` is not on PATH. A user without `uv` installed will see a message telling them their source file is missing and go looking in the wrong place. Either distinguish the two cases (catch `FileNotFoundError` from the Popen call separately) or include the underlying message in the error.

2. **Captured stderr is never drained, so ty can deadlock on a busy stderr.** In `src/uncoded/refs.py:64`, `subprocess.Popen` sets `stderr=subprocess.PIPE` and nothing reads from `proc.stderr`. If ty writes more than the pipe buffer (~64KB on Linux) to stderr during a session, it blocks on the write while `_run_exchange` blocks on stdout — classic subprocess deadlock. ty is usually quiet, so this is unlikely to bite often, but the failure mode is a silent hang with no diagnostic. Either redirect to `subprocess.DEVNULL` if you don't need stderr, or drain it in a background thread.

3. **`_uri_to_path` doesn't URL-decode and assumes a POSIX path layout.** `src/uncoded/refs.py:218`: `Path(urlparse(uri).path)`. URIs returned by ty (and produced by `Path.as_uri()`) percent-encode spaces and other characters; `urlparse(...).path` returns them raw. A repository in a path like `/Users/.../my project/...` will round-trip through `as_uri` as `file:///.../my%20project/...` and the resulting `Reference.rel_path` will contain a literal `%20`. On Windows the same function produces `/C:/...` which isn't a valid path. Use `urllib.parse.unquote` (and on Windows, strip the leading slash before the drive letter) to make this robust.

4. **LSP error path pays a 5-second cleanup tax.** In `src/uncoded/refs.py:171`, when `_run_exchange` raises on an LSP error response, the function exits before sending `shutdown`/`exit`. The `finally` in `query_references` then calls `_terminate`, which waits 5 seconds for the orphan ty process to exit on its own before calling `terminate()`. Every LSP error path costs that 5 seconds. Two cleaner shapes: send `exit` from the error branch before raising, or have `_terminate` call `proc.terminate()` immediately when the process is known to still be running.

5. **`_refs` catches bare `Exception`.** `src/uncoded/cli.py:189` collapses every remaining exception type to "Error: {e}" and exit 1. That hides genuine programming bugs (e.g. an `AttributeError` from a stale stub) behind a generic message. The only error this branch currently needs to handle is `RuntimeError` from `_run_exchange`; narrow the catch to that.

6. **`cast(ast.TypeAlias, node)` is a silent trap.** `src/uncoded/body.py:59`: the trailing branch in `resolve_name_position` assumes anything not matched by the preceding `isinstance` checks is a `TypeAlias`. That's true today, but if `_resolve_ast_node_from_source` ever returns a new node kind (for example because a future task adds another top-level shape), the cast quietly produces nonsense coordinates instead of failing. A defensive `raise UnsupportedNamePath` (or a dedicated assert) at the end is cheap insurance.

7. **`query_references` doesn't document its precondition.** `src/uncoded/refs.py:52` calls `in_path.as_uri()` via `_run_exchange`, which raises `ValueError` on a relative path. The CLI happens to pass an already-resolved path, but the docstring on this near-public function should say "in_path must be absolute" so a future caller doesn't trip on it.

## Nits

1. The README says results are "sorted by path then line" (`README.md` line 147 area), but `find_refs` sorts by `(rel_path, line, col)`. Mention column for accuracy.

2. The LSP error message renders the error dict via `repr` (`refs.py:172` — `f"ty LSP error: {response['error']}"`). Extracting `response["error"].get("message", ...)` would give the user a readable string instead of `{'code': -32602, 'message': 'not open'}`.

3. `_body` uses `Path(in_path)` while `_refs` uses `Path(in_path).resolve()` (cli.py lines 148 and 174). The asymmetry is load-bearing — `as_uri()` needs an absolute path — but a one-line comment explaining why would save the next reader a trip through `refs.py`.

alimanfoo:
Response to the review above.

## Non-blocking

1. **`uvx`-not-found error.** Addressed in 5c03815. `query_references` now wraps the `Popen` call and re-raises a `RuntimeError` with a clear "uvx not found on PATH — install uv" message; the source-file `FileNotFoundError` handler in `_refs` no longer catches it.
2. **stderr deadlock.** Addressed in dd284ac by redirecting `stderr=subprocess.DEVNULL` in the `Popen` call. uncoded doesn't need ty's stderr; discarding it eliminates the pipe-fill failure mode.
3. **`_uri_to_path` URL-decoding.** Addressed in c62043c with `urllib.parse.unquote` around the parsed path. Integration test added at a fixture path containing a literal space. Windows path handling held for a separate issue — current `as_uri()`/`urlparse` chain has the same blind spot today.
4. **5s cleanup tax on error path.** Addressed in dd284ac by moving the LSP error-raise after the shutdown/exit exchange. ty now receives a clean shutdown in both happy and error paths, so `_terminate`'s wait returns immediately.
5. **Narrow bare `Exception` catch.** Addressed in 5c03815 — `_refs` now catches `RuntimeError` specifically. Unexpected exception types propagate as tracebacks.
6. **`cast(ast.TypeAlias, node)` defensive raise.** Addressed in bbe6a59 — explicit `isinstance(node, ast.TypeAlias)` check followed by `raise UnsupportedNamePath` on unexpected node kinds.
7. **`query_references` precondition.** Initially documented in bbe6a59, then upgraded to a structural guard in 28fba0f — the function now raises `ValueError` at the boundary if `in_path` is not absolute, rather than letting the error surface from `Path.as_uri()` deep inside.

## Nits

1. **README sort accuracy.** Fixed in 3eeb415 — "sorted by path then line" → "sorted by path, then line, then column".
2. **LSP error message readability.** Fixed in 5c03815 — error message now extracts `error.get('message', error)` instead of rendering the full dict.
3. **`_body`/`_refs` asymmetry comment.** Added in 3eeb415 — two-line comment above `Path(in_path).resolve()` in `_refs` explains the `as_uri()` requirement.

src/uncoded/body.py
```diff
@@ -50,27 +105,26 @@ def resolve_body(name_path: str, in_path: Path) -> str:
...
             if name == head:
-                match = node
+                top_match = node
         elif isinstance(node, ast.TypeAlias) and tail is None and node.name.id == head:
-            match = node
+            top_match = node
 
-    if match is None:
+    if top_match is None:
         raise BodyNotFound(f"{name_path!r} not found in {in_path}")
```
alimanfoo: This exception is perhaps not named quite right any more, given that it can be raised also via find_refs. Perhaps SymbolNotFound?

alimanfoo: Agreed. Renamed in 2f3e7d0 — `BodyNotFound` → `SymbolNotFound` across body.py, cli.py, refs.py docstring, tests, namespace map, and stubs. `grep BodyNotFound` returns zero hits.

src/uncoded/body.py
```diff
@@ -18,29 +18,84 @@ class UnsupportedNamePath(Exception):
...
+    source = in_path.read_text()
+    node = _resolve_ast_node_from_source(
+        name_path=name_path, source=source, in_path=in_path
+    )
+    return _extract_body(node=node, lines=source.splitlines(keepends=True))
+
+
+def _resolve_ast_node_from_source(
+    *,
+    name_path: str,
```
alimanfoo: The name_path variable would seem to be a good candidate for the parse-don't-validate pattern, don't you think? Parse at the system boundary into a type, rather than validate all the way down here where the call stack bottoms out?

alimanfoo: Agreed — applied in 47ef798. `NamePath` is a `NamedTuple` with `head` + optional `tail`; `parse_name_path` is the system-boundary parser that raises `UnsupportedNamePath`. All public functions (`resolve_body`, `resolve_ast_node`, `resolve_name_position`, `find_refs`) now take `NamePath` directly. The CLI handlers parse at the entry; everything downstream is validated by construction. The segments-and-empty-check block in `_resolve_ast_node_from_source` is gone — the type carries the contract.

Follow-up 434060e dropped a now-dead `tmp_path` fixture from the `UnsupportedNamePath` tests (Junio caught it on audit).

alimanfoo (commented):
We have now several classes and functions defined in the body module but which are actually a common, core set required by both body and refs. Including NamePath, SymbolNotFound, parse_name_path. Should these live in a common module instead? We already have ast_helpers as a common module, perhaps this could be used, if so with a rename to match it's slightly broader scope?

Also, is there a reason why parse_name_path is a separate function and not the constructor for the NamePath class?

alimanfoo:
Two good questions; my take on each.

**Shared module for the common core.** Agreed the layering's off — `refs.py` reaches into `body.py` for `NamePath`, `parse_name_path`, `SymbolNotFound`, and the AST traversal machinery (`resolve_ast_node`, `resolve_name_position`, `_resolve_ast_node_from_source`, `_resolve_class_member`). Those are name-path-resolution surfaces, not body-retrieval surfaces. After the move, `body.py` shrinks to just `resolve_body` + `_extract_body` — actually about bodies.

On the destination: I'd lean toward a new dedicated module (something like `resolver.py` or `name_path.py`) rather than expanding `ast_helpers.py`. `ast_helpers` today is small node-inspection helpers (`property_kind`, `assign_target_name`); folding in the type, the parser, the exceptions, and the full traversal would change its character. A dedicated module communicates its purpose plainly and keeps `ast_helpers` at its current narrow scope. Open to the ast_helpers-with-rename shape if you prefer that — it's a judgement call about module breadth.

**`parse_name_path` vs constructor.** Mostly a constraint of the NamedTuple choice. NamedTuple fixes the constructor to `NamePath(head, tail)`, which is useful for direct construction in tests and elsewhere. To make `NamePath(s)` parse a string, we'd need to either:

- swap NamedTuple for a `@dataclass(frozen=True)` with `__post_init__` doing the parsing — loses NamedTuple's free destructuring (`head, tail = name_path`) but keeps everything else; or
- override `__new__` on the NamedTuple, which is doable but uglier than either alternative.

The closest equivalent that keeps NamedTuple is a `NamePath.parse(s)` classmethod — same shape as the free function, just namespaced. I'd actually prefer that over the standalone `parse_name_path` — better discoverability from the type, no extra module-level symbol.

So: my proposal would be to make both moves in one task — lift the resolver machinery to a new module, and convert `parse_name_path` to `NamePath.parse` while it's getting touched. Or split into two tasks if you'd rather review them separately. Your call.

alimanfoo:
Both moves landed in b7bee34, with a follow-on a75dad1 fixing the cross-module private import that the split exposed.

**Shared module.** `src/uncoded/resolver.py` is new and owns the name-path resolution surface: `NamePath`, `SymbolNotFound`, `UnsupportedNamePath`, `resolve_ast_node`, `resolve_name_position`, `resolve_ast_node_from_source`, and the private `_resolve_class_member`. `body.py` shrinks to 28 lines — `resolve_body` plus `_extract_body` — actually about bodies now. `refs.py` and `cli.py` import the common bits from `resolver`. Dependency graph: `resolver ← ast_helpers`; `body ← resolver`; `refs ← resolver`; `cli ← body + resolver + refs`. No cycles.

**`NamePath.parse` classmethod.** Free `parse_name_path` is gone. `NamePath.parse(s)` carries the same logic, namespaced on the type. CLI callsites are now `NamePath.parse(name_path)` at the boundary. Couldn't make `NamePath(s)` the parser without giving up the NamedTuple shape (constructor signature is fixed to `(head, tail)`), and the NamedTuple machinery — free destructuring, equality, hashability — was worth keeping.

**Follow-on a75dad1.** The split surfaced a cross-module private import: `body.py` reached into `resolver.py` for `_resolve_ast_node_from_source` (the shared single-read helper from earlier in the session). Promoted to public as `resolve_ast_node_from_source` with a docstring documenting why it exists (lets `resolve_ast_node` and `resolve_body` share one file read). `_resolve_class_member` stays private — it's only called from within `resolver.py`.

PR #149, opened 2026-05-17
[claude] tidy refs.py: pin rationale, private API, gated tests

### LINKED ISSUE #146: [claude] TestFindRefs and TestQueryReferences spawn ty on every test
\`tests/test_refs.py:TestFindRefs\` and \`TestQueryReferences\` invoke the real ty LSP via \`uvx --from ty=={TY_VERSION} ty server\` on every test run. Each test pays ~1–2s of subprocess setup; the full suite slows accordingly and fails opaquely on machines without \`uv\` or without network access.

The exchange-level branches are already covered by \`TestRunExchange\` (which feeds a mock \`BytesIO\` stream), so the real-subprocess tests defend the wire-level integration with ty rather than uncoded's own branching logic.

Suggested direction:

- Mark the real-subprocess tests with a pytest marker (e.g. \`@pytest.mark.integration\`), and gate them behind a config switch so the default suite runs fast.
- Or extend \`TestRunExchange\`-style fixture coverage so the integration tests can be replaced entirely, leaving one or two real-subprocess smoke tests gated behind the marker.

Either way the fast suite stays fast and a deliberate run exercises the wire protocol.

### LINKED ISSUE #147: [claude] TY_VERSION pin has no rationale beside it
\`src/uncoded/refs.py:15\` declares \`TY_VERSION = \"0.0.37\"\` as a bare constant. The rationale for the pin — ty is pre-1.0 with known \`textDocument/references\` edge cases (#111 lists astral-sh/ty #3406, #3387, #3369, #3411, #3410) — lives only in the PR history that introduced it.

A maintainer bumping ty in the future will see the constant without that context and may either pin too liberally (\`>=0.0.37\`) or remove the pin once ty looks stable, missing the references-API edge-case context.

Suggested direction: a one-line comment at the constant naming the reason for the strict pin, with a pointer to #111 (or wherever the bump policy lives) so the next person to touch the constant has the context.

### LINKED ISSUE #148: [claude] query_references is public but returns a private _LSPLocation
\`src/uncoded/refs.py:query_references\` is a public function (no underscore prefix) with return type \`list[_LSPLocation]\` — a private dataclass. A public function with a private return type is an inconsistency: external callers can reach it, but the type name signals "internal detail, may change."

\`find_refs\` is the intended entry point for callers: it takes care of position discovery, calls \`query_references\` underneath, and converts the raw LSP locations (0-indexed, absolute path) into the public \`Reference\` shape (1-indexed, relative path, sorted). Calling \`query_references\` directly leaves the caller with raw LSP coordinates they then have to convert.

Suggested direction: make \`query_references\` private — rename to \`_query_references\`. The test suite already imports several private helpers from \`refs.py\` directly, so test access isn't a constraint. Promoting \`_LSPLocation\` is the inverse option but probably wrong: the raw shape isn't a contract callers should rely on.

alimanfoo:
Three small coherence gaps surfaced on `src/uncoded/refs.py` once it settled in after #111:

- The `TY_VERSION` constant sat bare. The rationale for the strict pin — ty is pre-1.0 with known `textDocument/references` edge cases — lived only in the #111 history, leaving a future maintainer no context at the call site. A two-line comment now points back to #111. Closes #147.
- `query_references` was public-named but returned the private `_LSPLocation`. Its only non-test caller was `find_refs` in the same module. Renamed to `_query_references` and converted to keyword-only to match every other private helper in the file. `find_refs` stays the single public entry point. Closes #148.
- Eight tests spawned the real `ty` subprocess on every `pytest` run — about ten seconds of suite time and a hard dependency on `uvx` and network. They're now decorated `@pytest.mark.integration` and deselected by default. A new unit test mocks `subprocess.Popen` to cover the success branch of `_query_references` so the default suite stays at 100% branch coverage without any pragma. CI runs both buckets via `pytest -m "integration or not integration"`. Closes #146.

The renamed CLI comment and the `ValueError` message both now cite `_query_references` rather than the old name.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,4 +30,4 @@ jobs:
       - uses: astral-sh/setup-uv@v6
       - run: uv python install ${{ matrix.python-version }}
       - run: uv sync --extra dev
-      - run: uv run pytest -v
+      - run: uv run pytest -v -m "integration or not integration"
--- a/.uncoded/namespace.yaml
+++ b/.uncoded/namespace.yaml
@@ -66,7 +66,7 @@ src/:
         line:
         character:
       find_refs:
-      query_references:
+      _query_references:
       _find_root:
       _terminate:
       _to_rel_path:
--- a/.uncoded/namespace.yaml
+++ b/.uncoded/namespace.yaml
@@ -349,6 +349,7 @@ tests/:
       test_uvx_not_found_raises_runtime_error:
       test_relative_path_raises_value_error:
       test_returns_empty_list_when_no_references:
+      test_returns_lsp_locations_when_popen_succeeds:
     TestRunExchange:
       test_lsp_error_raises:
       test_empty_result_list_returns_empty:
--- a/.uncoded/stubs/src/uncoded/refs.pyi
+++ b/.uncoded/stubs/src/uncoded/refs.pyi
@@ -16,7 +16,7 @@ TY_VERSION = '0.0.37'
 def find_refs(name_path: NamePath, in_path: Path) -> list[Reference]:
     ...
 
-def query_references(in_path: Path, position: tuple[int, int]) -> list[_LSPLocation]:
+def _query_references(*, in_path: Path, position: tuple[int, int]) -> list[_LSPLocation]:
     ...
 
 def _find_root(in_path: Path) -> Path:
--- a/.uncoded/stubs/tests/test_refs.pyi
+++ b/.uncoded/stubs/tests/test_refs.pyi
@@ -7,7 +7,7 @@ import textwrap
 from pathlib import Path
 from unittest import mock
 import pytest
-from uncoded.refs import Reference, _find_root, _LSPLocation, _read_message, _read_response, _run_exchange, _terminate, _to_rel_path, find_refs, query_references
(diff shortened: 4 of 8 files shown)


alimanfoo:

## Nits
1. `tests/test_refs.py:44-122` — `@pytest.mark.integration` is applied per-method to all six `TestFindRefs` tests. Since the entire class is integration-only, a class-level `pytestmark = pytest.mark.integration` (or decorating the class) is more concise and harder to forget when a new method is added later. `TestQueryReferences` mixes integration and unit tests, so the per-method form there has to stay.
2. `src/uncoded/refs.py:66` — the `ValueError` message now reads `"_query_references requires an absolute path; got ..."`. The underscore-prefixed name appears in user-visible output, which reads oddly given the function is now explicitly internal and callers go through `find_refs`. Since the PR has touched this string anyway, consider phrasing it without the function name, e.g. `"absolute path required; got ..."`. The existing test matches on `"absolute path"`, so a reword wouldn't disturb it.

alimanfoo:
Both nits accepted and addressed.

1. **Class-level marker on `TestFindRefs`** — addressed in `11f8dc2`. Six per-method decorators collapsed to a single `@pytest.mark.integration` on the class. `TestQueryReferences` left as-is since it mixes integration and unit tests.
2. **Function name in `ValueError`** — addressed in `1982c5e`. Message now reads `f"absolute path required; got {in_path!r}"`. The existing `test_relative_path_raises_value_error` match on `"absolute path"` still passes unchanged.

PR #151, opened 2026-05-17
[claude] complete migration from Serena/MCP to uncoded body + refs

### LINKED ISSUE #46: [claude] find_referencing_symbols returned incomplete results after structural move

### LINKED ISSUE #111: [claude] proposal: replace Serena/MCP with uncoded body + refs CLIs (cross-platform)

### LINKED ISSUE #120: [claude] find_referencing_symbols misses kwarg-bound writes to dataclass fields
`find_referencing_symbols` from the Serena tools — recommended by the dispatch rule in `CLAUDE.md` for tasks like rename and dead-code checks — appears to miss keyword-argument writes that bind to dataclass fields. 

During plan work for #116, calling it on `StubAssignment.is_type_alias` returned 5 sites (the declaration, one attribute read in `_format_assignment_body`, three attribute reads in tests). A plain `grep -rn is_type_alias` returned 9: the same 5 plus three kwarg writes (`_extract_assignment`, `_property_attribute`, `test_type_alias_pep695_rendered`) and the generated stub.

Plausible mechanism: the language server resolves attribute accesses (`obj.field`) and identifier references to the field, but classifies keyword-argument names in dataclass constructor calls as parameter-name tokens of the synthesised `__init__` — not as references to the field itself.

The dispatch rule currently presents `find_referencing_symbols` as the complete answer for "Find callers, or check whether a symbol is dead." For dataclass field renames or deletions, that's not safe alone. Suggested direction: either add a caveat to the dispatch rule (recommending a `grep` cross-check on field renames), or check whether the limitation can be addressed in `serena_setup` configuration.

alimanfoo:
This is step 5 of #111 — completing the migration away from Serena/MCP for the cross-symbol operations (rename, edit-by-symbol, safe-delete). Steps 1–4 of that plan are already in. After this PR, agents using uncoded read symbol bodies with `uncoded body`, find references with `uncoded refs`, and edit using the platform's native `Edit` tool with `uncoded body`'s output as the `old_string`. No language-server layer is needed for any of it.

Removing Serena also resolves the staleness class of bugs that motivated this whole thread — #46 and #120 — by removing the long-lived LSP buffer where the staleness lived. Each `uncoded refs` invocation runs against a fresh subprocess that reads from disk, so the bug has no place to hide anymore.

The dispatch-rule prose injected into `CLAUDE.md`/`AGENTS.md` is rewritten in place: the orient/understand/act spine stays, step 3 is recast around `uncoded body` + `uncoded refs` + `Edit`, and the four Serena tool bullets collapse into a single "Edit a symbol" bullet that names the patterns (single-symbol edit, rename-across-codebase, safe-delete) in terms of those three calls. The Serena configuration plumbing — `uncoded setup`, `serena_setup` module, dogfood `.mcp.json` / `.serena/` / `.claude/settings.json`, `.gitignore` and pre-commit hook entries — is all removed.

Closes #46
Closes #111
Closes #120


alimanfoo:
**Recommendation:** blocking concern below, plus a couple of non-blocking points and nits.

## Blocking

1. `pyproject.toml:8` — the `description` field still reads *"Symbol indexes and language-server setup for AI-assisted code navigation"*. This PR removes the language-server-setup half of that promise, and the field is what ships to PyPI on the next release. Bring it in line with the new README intro (which now describes `uncoded body` + `uncoded refs` instead).

## Non-blocking

1. `src/uncoded/config.py:22` — `read_project_name` has no production callers after `serena_setup.py` was deleted; only `tests/test_config.py` exercises it. It was added for `setup()` (per the GH68 thread cited in the now-deleted Serena tests). This is consequential adjacency from this PR rather than a pre-existing fossil — the same edit that dropped `serena_setup` orphaned this function.

2. `src/uncoded/instruction_files.py:102–107` — the new "Edit a symbol" bullet packs three workflows (single-symbol edit, rename across codebase, safe-delete) into one dense paragraph. Previously each had its own heading. An agent scanning for "how do I safely delete" no longer has a heading to land on and has to read the bullet through. Splitting back into three short bullets would restore scanability.

## Nits

1. `README.md:155` — item 5 in "How agents use it" packs the same three workflows into one long semicolon-separated sentence. Items 1–4 are short, parallel bullets; item 5 breaks the rhythm. Splitting it (or sub-bulleting) would mirror items 1–4.

2. `src/uncoded/instruction_files.py:41–43` — the third "rule firing" example breaks the parallel with the first two. The first two end "that is \`uncoded body\`" / "that is \`uncoded refs\`" (single tool). The third ends "that is \`uncoded refs\` to confirm dead, then \`Edit\`" (two-step workflow). Reads slightly off; minor.

3. `src/uncoded/instruction_files.py:43–44` — short-line wrap at "matches comments," looks like a leftover from preserving the old break while substituting a longer phrase. Cosmetic only; Markdown renders fine.

alimanfoo:
Thanks for the review. Triage and follow-up:

**Blocking**

1. *pyproject description* — accepted. Addressed in 508fb7d ("update package description for the body+refs cutover").

**Non-blocking**

1. *read_project_name orphaned* — accepted. Verified zero production callers via `uncoded refs` (four test-only references). Removed function + tests in 50e9d01 ("remove orphaned read_project_name and its tests"). `tomllib` stays — still used by `read_source_roots` and `read_instruction_files`.

2. *"Edit a symbol" bullet density* — accepted. Split back into three peer bullets (Edit / Rename / Safely delete) in 7622022 ("split 'Edit a symbol' bullet for scanability").

**Nits**

1. *README item 5 rhythm* — accepted. Split into three top-level items (5/6/7) mirroring items 1–4 in 6cc1e84 ("split README item 5 into edit/rename/safe-delete items").

2. *Third "rule firing" example breaks parallel* — rejected. The third case genuinely *is* a two-step workflow rather than a single tool; dropping the "to confirm dead, then `Edit`" trailer to restore parallel would lose information an agent reading the example actually needs. Keeping the trailer.

3. *Short-line wrap leftover at "matches comments,"* — accepted. Reflowed alongside the bullet split in 7622022.

src/uncoded/instruction_files.py
```diff
@@ -22,10 +22,9 @@
 ## How to read and edit code in this codebase
 
 This repo uses [uncoded](https://github.com/alimanfoo/uncoded) to maintain
-a symbol index over its source code, with three tools over that index:
-`uncoded body` for reading a symbol's body, `uncoded refs` for finding
-references, and [Serena](https://github.com/oraios/serena) for cross-symbol
-operations — renaming, editing by symbol, and safe-deleting.
+a symbol index over its source code, with two tools over that index:
+`uncoded body` for reading a symbol's body and `uncoded refs` for finding
+references.
```
alimanfoo: "two tools over that index:" -- this isn't strictly correct, as the tools don't operate over the namespace index or the stubs. Perhaps just "two associated tools:".

AGENTS.md
```diff
@@ -44,12 +43,12 @@ uv run uncoded sync
...
 
-# Generate Serena + ty MCP and Claude Code configuration
-uv run uncoded setup
-
 # Print the source body of a named symbol to stdout
 uv run uncoded body <name_path> --in <relative_path>
 
+# Find references to a symbol
+uv run uncoded refs <name_path> --in <relative_path>
+
```
alimanfoo, line 51: Just noticing that often uncoded will be added as a dev dependency in a host repo, but not as a production dependency. So `uv run` may sometimes not work? Would using `uvx` for the examples be more robust? Perhaps worth raising as a follow-on issue.

pyproject.toml
```diff
@@ -5,7 +5,7 @@ build-backend = "hatchling.build"
 [project]
 name = "uncoded"
 dynamic = ["version"]
-description = "Symbol indexes and language-server setup for AI-assisted code navigation"
+description = "Symbol index with `uncoded body` and `uncoded refs` for AI coding agent navigation"
```
alimanfoo:
```suggestion
description = "Symbol index with associated code reading tools for AI coding agent navigation"
```

Suggest to generalise a little.

README.md
```diff
@@ -166,12 +152,13 @@ Agents following that protocol:
 2. Read the relevant `.pyi` stubs to understand imports, signatures, constants, and class members.
 3. Run `uncoded body <name_path> --in <relative_path>` when they need implementation detail for a specific symbol.
 4. Run `uncoded refs <name_path> --in <relative_path>` for impact analysis — callers, dead-symbol checks. See [Find references to a symbol](#find-references-to-a-symbol) for detail.
-5. Use [Serena](https://github.com/oraios/serena) for cross-symbol operations — renaming, editing by symbol, and safe-deleting. See [Using uncoded with a language server](#using-uncoded-with-a-language-server) for setup.
+5. Edit a symbol using `Edit` with `uncoded body`'s output as `old_string`.
+6. Rename across the codebase using `uncoded refs` to enumerate every site, then `Edit` at each.
+7. Safely delete by running `uncoded refs` first — the output must be empty — then `Edit` to remove.
 
 The split is deliberate: `uncoded` provides a stable map and signature index;
 `uncoded body` resolves the current source body; `uncoded refs` maps call
```
alimanfoo: Noticing the word "current" here is not so obvious. Perhaps "resolves a symbol's source body"?

alimanfoo:
Thanks for the review. Triage and follow-up:

1. *"two tools over that index"* (`instruction_files.py:27`) — accepted. Changed to "two associated tools" in 296c56b. AGENTS.md/CLAUDE.md picked up via the same commit's `uncoded sync` regen.

2. *`uv run` vs `uvx` for examples* (`AGENTS.md:51`) — out of scope as you suggested; held for post-merge triage.

3. *Generalise pyproject description* (`pyproject.toml:8`) — accepted. Applied the suggested wording verbatim in 6f5f1a6.

4. *"the current source body"* (`README.md:160`) — accepted. Changed to "a symbol's source body" in dede8f6.

PR #156, opened 2026-05-18
tidy refs path normalization, extract dispatch-rule, adopt uvx

### LINKED ISSUE #150: _refs comment cites the wrong mechanism for the relative-path ValueError
Noticed during #149, the comment in `src/uncoded/cli.py:_refs` justifying the `.resolve()` call cites a mechanism that isn't on the relative-path code path:

```python
# resolve() is required: _query_references calls Path.as_uri(), which
# raises ValueError on a relative path.
target = Path(in_path).resolve()
```

`_query_references` does not call `Path.as_uri()` directly — `_run_exchange` does, two levels down. And on a relative `in_path`, the raise comes from the explicit `if not in_path.is_absolute()` guard at the top of `_query_references`, so `as_uri()` is never reached on that path.

The underlying reason for `.resolve()` is correct (an absolute path is required downstream); only the cited mechanism is wrong. A future maintainer reading the comment gets the wrong mental model of where the constraint lives.

Suggested direction: trim the imprecise mechanism clause and keep the bare why — something like `# resolve() is required: _query_references rejects relative paths.` Removing the clause is cleaner than correcting it.

### LINKED ISSUE #152: Remove out-of-date Serena content in lsp-research.md and .uncoded/reviews/
Noticed during #151, when post-merge sweep flagged that two historical documents still position Serena as current tooling, even though #151 just completed the migration away from Serena/MCP.

Both files predate the resolution of #111 and were not updated by it:

- `resources/lsp-research.md` — a research survey of the LSP-over-MCP landscape that informed the original Serena adoption. Serena is named extensively as the leading example, with no preface noting the project ultimately took a different path (the body+refs CLIs shipped in #111 instead).
- `.uncoded/reviews/2026-04-25-001215.md` — a timestamped coherence-review report from a past sweep, with several references to `serena_setup.py` and Serena tooling. The module it cites is now deleted.

A future reader picking up either file could reasonably assume Serena is still in use, or still the recommended direction. Both noticed independently by reviewers post-merge.

Suggested direction: remove both. Neither has a current consumer (the research's conclusion was reached and shipped; the review report's observations were either acted on or are no longer current). If `lsp-research.md` is worth keeping for archival reasons, an archive marker or preface noting its historical status would be a smaller alternative.

### LINKED ISSUE #153: `_SECTION_BODY` Markdown lives as a Python string literal in instruction_files.py
Noticed during #151. `_SECTION_BODY` in `src/uncoded/instruction_files.py` is ~100 lines of agent-facing Markdown embedded as a triple-quoted Python string. After the prose rewrite in #151 it's by far the largest thing in the module, and every edit lands inside Python rather than Markdown tooling — no preview, no link-checking, no Markdown lint, and the surrounding Python infrastructure (ruff format, ty) provides no useful signal on the content.

Consumers are contributors editing the dispatch-rule prose; the prose itself is generated into `CLAUDE.md`/`AGENTS.md` by `sync_instruction_file`, so a structural move doesn't change the agent-facing output.

Suggested direction: extract the body to a sibling `.md` template loaded at module import via `importlib.resources` (or similar). The Python module becomes the small wiring; the Markdown lives in its native form.

### LINKED ISSUE #154: AGENTS.md says `uncoded refs` lists "all call sites"
Noticed during #151. AGENTS.md's Approach paragraph (manually maintained, above the `<!-- uncoded:start -->` markers) reads:

> "Alongside the index, uncoded ships `uncoded body` to read symbol bodies and `uncoded refs` to list all call sites."

`uncoded refs` finds **all references** to a symbol, not only call sites — imports, attribute access, type annotations, and other reference shapes all appear in the output. Describing it as "all call sites" undersells the tool and could lead an agent to skip running it before a rename (thinking it covers only method/function calls).

Suggested direction: change "to list all call sites" to language that names the broader set, e.g. "to find all references" or "to list every reference". The injected dispatch-rule section already describes the tool accurately; this is a one-paragraph fix in the manually maintained portion.

### LINKED ISSUE #155: Consider `uvx` over `uv run` for AGENTS.md command examples
Noticed during #151 PR review. AGENTS.md's Commands section lists example invocations as `uv run uncoded sync`, `uv run uncoded body …`, and so on. `uv run` requires uncoded to be a project dependency in the host repo — but uncoded is often added as a *dev* dependency rather than a production one, and may not be a dependency at all in repos that consume the index without re-syncing.

`uvx` runs the published package ephemerally from PyPI, independent of the host project's dependency configuration. That works in any host repo with `uv` installed, including the ones where `uv run uncoded …` doesn't resolve.

Suggested direction: replace `uv run uncoded …` with `uvx uncoded …` in the AGENTS.md command examples, or document both forms with a note on when each applies. Worth checking whether the same change applies to README.md examples.

alimanfoo:
This branch bundles five small maintenance items. They share a common theme — a few places where prose lived where it shouldn't, where the explanation didn't match the code, or where invocation guidance was inconsistent. Cleaning them up together keeps the diff coherent.

## What's in this PR

**Path normalization for `uncoded refs`** (#150). The absolute-path requirement for the LSP exchange comes from `Path.as_uri()` in `_run_exchange`. It was previously enforced two layers above (a guard in `_query_references`) and pre-emptively satisfied three layers above (`.resolve()` in `_refs`), with a comment in `_refs` explaining why that cited the wrong layer. Normalization now lives at the function that actually imposes the constraint. The misleading comment, the guard, and the defensive `.resolve()` are all gone. `find_refs`'s observable contract widens — relative paths no longer raise — but no existing caller passes relative paths.

**Out-of-date Serena documents removed** (#152). `resources/lsp-research.md` and `.uncoded/reviews/2026-04-25-001215.md` pre-dated the Serena removal in #151 and referenced now-deleted modules. Both gone, along with their (then-empty) parent directories.

**Dispatch-rule Markdown moves out of the Python module** (#153). The ~105-line agent-facing dispatch-rule body lived as a triple-quoted Python string inside `src/uncoded/instruction_files.py`. It now lives in `src/uncoded/dispatch_rule.md` and is loaded at import via `importlib.resources`. Future edits to the prose happen in Markdown, not Python. Hatchling's default wheel target ships the `.md` alongside the `.py` files without any pyproject change.

**`uncoded refs` framings broadened** (#154). Five hand-maintained surfaces in AGENTS.md and README.md described the tool as returning "call sites" or "callers" — but it returns all references (imports, type annotations, attribute access, calls). The wording now names "references" first; callers, dead-symbol checks, and pre-rename footprint stay as example uses.

**Single `uvx` invocation form** (#155). All `uncoded` command examples in AGENTS.md and README.md now use `uvx uncoded …`. README's Install section drops `pip install uncoded` and `uv add uncoded` — running via `uvx` needs no install. Project-tooling commands (`pytest`, `pre-commit`) keep `uv run` — that's contributor workflow, not host-repo use. The dispatch-rule prose's own command examples were updated to match in #153's commit, so generated host-repo instructions stay consistent.

Closes #150
Closes #152
Closes #153
Closes #154
Closes #155


alimanfoo:

## Nits

1. `AGENTS.md`/`CLAUDE.md` line 35 — the new "Commands" intro is a single unwrapped long line, where the rest of the file wraps to ~70 columns. Wrapping it would match the surrounding style and keep future diffs in this region clean.

2. `README.md` line 38 — "install uv from the link if you don't have it" reads awkwardly because "the link" points back at the parenthetical Markdown link immediately before it. Naming uv directly (e.g. "install uv from https://docs.astral.sh/uv/ if you don't already have it") removes the stumble.

3. `README.md` "Install" section header — the section now says no install is needed. The heading could be renamed (e.g. "Run via uv") or folded into the next section so the header matches the content.

## Out of scope but noticed

1. The dispatch-rule prose in `src/uncoded/dispatch_rule.md` still leads Step 3 with "use `uncoded refs` to find callers" (line 60) and uses "Find callers, or check whether a symbol is dead" as a bullet header (line 73). The very top of the file correctly names the tool's contract ("finding references"), so the action-oriented "callers" framing further down is defensible — but it's now slightly out of step with the broadening campaign elsewhere. Issue #154 explicitly excluded this prose; flagging here only because someone reading the file end-to-end might notice the mixed framing.

2. `.pre-commit-config.yaml` line 27 keeps `uv run uncoded sync`, while the README's pre-commit example shows `uvx uncoded sync`. The divergence is correct (this repo develops uncoded, so the editable install is the right target), but the rationale isn't recorded in the file itself. A one-line comment would help future contributors who notice the mismatch.

alimanfoo:
**Nits (all accepted, addressed in 74a4159):**

1. **AGENTS.md intro line wrap** — reflowed to three ~74-col lines matching surrounding style.
2. **README "install uv from the link"** — reworded to "Install uv if you don't already have it"; the parenthetical Markdown link to https://docs.astral.sh/uv/ now stands on its own without backward reference.
3. **README `## Install` header** — renamed to `## Install uv`. The heading now names what you might actually need to install (uv itself), flowing naturally into `## Configure` and `## Use`.

**Out of scope but noticed (held for post-merge triage):**

1. **Dispatch-rule prose "callers" framings inside `dispatch_rule.md`** — held. The inner prose is pre-existing and was explicitly excluded from #154's scope. Will be considered in post-merge triage.
2. **`.pre-commit-config.yaml` rationale comment** — held. The divergence from the README's pre-commit example is intentional (contributor vs. host-repo use) but the rationale isn't documented in the file itself. Will be considered in post-merge triage.

PR #165, opened 2026-06-17
Add Markdown doc-orientation index

### LINKED ISSUE #164: Add analogous functionality for markdown
## Problem

uncoded gives agents a top-down map of *code* so they navigate by structure
instead of grepping and skimming. Prose docs have the same failure mode:
agents skim a long README or design doc and fill gaps from pretraining.
Give docs the same orientation layer.

## Scope (suggested)

An *outline only* — no `body`/`refs`/stubs for Markdown. Rationale: a Markdown
heading is literal, greppable text, so once an agent sees the outline it jumps
to the section with an ordinary Read/grep — the thing grep is actually good at.
The index's job for prose is purely orientation. This stays true to the
dispatch rule (symbol name → index; free text → grep).

## Config (`[tool.uncoded]`)

New optional `doc_roots`, mirroring `source_roots` — a list of directories
(walked for `*.md`) or individual `.md` files. Opt-in: absent/empty ⇒ no
Markdown indexing (fully backward-compatible).

## Config home — language-neutral

Discovery walks up for `pyproject.toml` *or* `.uncoded.toml`. Python repos
keep `[tool.uncoded]` in `pyproject.toml` (no change). Non-Python repos
(e.g. a Claude plugin marketplace — JSON manifests, skill `.md` files, READMEs,
no Python) use `.uncoded.toml` at the root, so there's no false "this is a
Python project" signal. If both exist, `pyproject.toml` wins. Config never
lives inside the generated `.uncoded/` dir (that dir is all do-not-edit
output).

## Artefact — `.uncoded/docs.yaml`

A separate file, fully independent of `namespace.yaml`. Same "pure key
hierarchy, no values" style, with its own header comment. A filesystem tree of
`doc_roots`, each `.md` file nesting its heading hierarchy by level:

```yaml
docs/:
  guide.md:
    Getting started:
      Install:
      Configure:
README.md:
  uncoded:
    Problem:
    Approach:
    Commands:
```

## `sync` / `check`

Per-root-type and independent:

- `source_roots` present ⇒ build `namespace.yaml` + stubs + code instruction
  section.
- `doc_roots` present ⇒ build `docs.yaml` + docs instruction section.

Require at least one root-type; error ("nothing to index") if neither. Each
artefact is emitted only for its configured roots.

## Instruction sections

Split the single injected section into two independently-managed,
marker-delimited sections in each instruction file — a code-navigation section
and a docs-navigation section — each present only when its roots are
configured. The docs section is short: read `docs.yaml` first to orient;
heading text is literal — Read/grep to the section. A docs-only repo therefore
gets no Python dispatch-rule noise.

## Open considerations (not yet decided)

- Duplicate sibling headings collide as YAML keys — need a tie-break (suffix
  counter? keep first? list form?).
- Ignore `#` lines inside fenced code blocks; decide whether to support Setext
  (`===`/`---`) headings or ATX only.
- Heading text with YAML-special characters relies on the dumper's quoting
  (already handled for code).

alimanfoo:
This PR adds a Markdown doc-orientation index to uncoded, addressing #164. uncoded already gives agents a top-down map of code; this extends the same orientation layer to prose docs, and makes the config language-neutral so non-Python repos can use it.

## Session type

Enhancement.

## Consumers

1. An AI coding agent in an uncoded-enabled repo. Today it has a code map but nothing for prose, so it skims long Markdown files and fills gaps from pretraining. It needs an outline to orient, then jump to the right section.
2. A maintainer configuring uncoded in a Python repo, who wants to opt into doc indexing alongside the code index, with no change if they don't.
3. A maintainer of a non-Python repo (for example a Claude plugin marketplace: JSON manifests, skill `.md` files, READMEs, no Python), who wants doc orientation with no false "this is a Python project" signal and no Python dispatch-rule noise.

## Use cases

- An agent loads `.uncoded/docs.yaml` at task start, sees the heading hierarchy of every configured Markdown file, and navigates to the right section, then Reads or greps to the literal heading.
- A maintainer adds `doc-roots` to config and runs `uvx uncoded sync`, and gets `.uncoded/docs.yaml` plus a docs-navigation section injected into the instruction files.
- A maintainer of a non-Python repo puts uncoded config in `.uncoded.toml` with only `doc-roots`, runs sync, and gets doc indexing with no Python artefacts and no code-dispatch-rule section.
- A maintainer runs `uvx uncoded check` in CI and gets a non-zero exit when `docs.yaml` or the docs instruction section is stale.

## Constraints

- Backward compatible: an absent or empty `doc-roots` means no Markdown indexing, and existing Python-only repos behave exactly as now.
- Outline only: no `body`, `refs`, or stubs for Markdown. The index's job for prose is purely orientation; headings are literal greppable text.
- `docs.yaml` is independent of `namespace.yaml`: a separate file, same pure-key-hierarchy style, with its own header comment.
- Per-root-type independence: `source-roots` present builds the code artefacts (map, stubs, code instruction section); `doc-roots` present builds the doc artefacts (`docs.yaml`, docs instruction section). At least one root type is required, otherwise sync errors with "nothing to index".
- Config never lives inside `.uncoded/`; that directory is all generated do-not-edit output.
- Config home: discovery walks up for `pyproject.toml` or `.uncoded.toml`. A `pyproject.toml` wins only when it carries a `[tool.uncoded]` section; a bare `pyproject.toml` does not shadow a sibling `.uncoded.toml`. A directory holding both a `[tool.uncoded]` `pyproject.toml` and a `.uncoded.toml` is a configuration error. Across directories, the nearer file wins.
- ATX headings only (`#`, `##`, and so on); Setext headings (text underlined with `===` or `---`) do not count. A `#` line inside a fenced code block is not a heading.

Duplicate sibling heading handling is left to the design phase; the requirements do not constrain it.

## System non-goals

- uncoded will never provide `body`, `refs`, or stubs for Markdown. Prose orientation is outline-only by design: a Markdown heading is literal greppable text, so once an agent sees the outline it jumps with an ordinary Read or grep.

Closes #164


alimanfoo:
# Code Analysis

A read of where Markdown doc indexing would land and what it touches. The generation pipeline is a clean three-stage pattern repeated per artefact, so most of this work is parallel structure; two surfaces need real reshaping.

## The orchestration spine — `cli._sync`

One sequential function drives everything: find project root, read source roots, validate each root is a directory, render and write the namespace map, build stubs per root, sync each instruction file, sync the skill, then account for changes. Every writer returns a "changed" signal that `_sync` sums; in check mode a non-zero sum returns 1. This is the spine the doc branch plugs into, and where the "require at least one root type, else error 'nothing to index'" rule must live. Today nothing enforces that, because `source-roots` is mandatory.

## Config layer — `config.py`

Three functions, two interface shapes. `find_pyproject_toml(start)` walks up for `pyproject.toml` only. `read_source_roots(pyproject_path)` takes a resolved file path, opens it, and reads `tool.uncoded.source-roots`, raising `LookupError` if the section is absent. `read_instruction_files(start)` takes a start directory, re-walks via `find_pyproject_toml`, reads `tool.uncoded.instruction-files`, and falls back to a default list.

Two structural notes. First, the config location — the `tool.uncoded` section path plus the open-load-read sequence — is written out in both readers. That is one fact (where uncoded config lives) in two places; it will drift the moment `.uncoded.toml` introduces a second home, where keys may sit at top level rather than under `[tool.uncoded]` (a design call). Second, the two readers have inconsistent interfaces: one takes a resolved path, the other a start directory. Adding `doc-roots` and the `.uncoded.toml` home touches all of this; a single "find the config home once, read all keys" loader is the natural shape.

Naming: TOML keys are kebab-case (`source-roots`, `instruction-files`) mapped to lists of paths. The issue's `doc_roots` would read as `doc-roots` to match the existing convention.

## Project-root discovery is hardcoded to Python — `cli._find_project_root`

The walk, the returned root (the parent of `pyproject.toml`), and the error text ("Create one with a `[tool.uncoded] source-roots` entry") all name `pyproject.toml`. The language-neutral goal has to unwind this into "find the config home (either file; `pyproject.toml` wins), derive the root from it".

## Artefact build and render pattern — `namespace_map.py`

`build_map(modules)` produces a pure nested dict with `None` leaves; `render_map(dict)` runs `yaml.dump` with a clean dumper (clean indentation, null rendering) and a prepended header, with sort disabled to preserve insertion order. `docs.yaml` mirrors this exactly: a builder making a `directory, file, heading, subheading` dict, and a renderer with its own header. The clean dumper and the null rendering are reusable; the header is namespace-specific. The "pure key hierarchy, no values" style holds only by the builder emitting `None` leaves plus the dumper config — nothing checks that `docs.yaml` follows the same style.

## Source walk — `extract.iter_source_files`

Walks every `*.py` under the root in sorted order, parses each via the AST, skips syntax errors with one stderr warning, and yields source plus relative path. The Markdown analogue differs two ways: `doc-roots` allows an individual `.md` file as well as a directory, so the walker handles both; and heading extraction has no AST — it is a line scanner for ATX headings that must skip `#` lines inside fenced code blocks. This heading parser is the one genuinely new mechanism; the rest is parallel structure.

## Instruction-file injection is single-section — `instruction_files.py`

One marker pair (`<!-- uncoded:start -->` and `<!-- uncoded:end -->`). The injected section wraps `dispatch_rule.md`, loaded via `importlib.resources` (the convention since #153 and #156). `_replace_or_append(existing, section)` finds the one pair and replaces it, else appends. The requirement to carry two independently-managed sections (a code-navigation section and a docs-navigation section, each present only when its root type is configured, and removed when it is not) does not fit this model. The single-pair `_replace_or_append` does not generalise to two distinct markers, and there is no path today to remove a section whose root type went away. This is the biggest reshaping in the work.

## Leaf I/O — `sync.py`

`sync_file` writes if content differs, is check-aware, anchors on the project root, and returns a changed flag. `remove_file` is the parallel removal path. Every artefact, and any removed section, routes through these unchanged. `skill.py`'s `sync_skill` is a precedent for emitting content to fixed paths with check mode.

## Architecture and conventions the work touches

- Three-stage pipeline per artefact: build (data), render (string), sync_file (diff and write, check-aware). Clean separation; no check enforces it.
- Single project-root anchor; all paths project-relative; log lines name the given path.
- The check flag threads as a bool through every writer; `_sync` sums the change signals.
- Injected prose lives in `.md` resources loaded via `importlib.resources`.
- Config lives under `[tool.uncoded]` in `pyproject.toml`, kebab-case keys mapped to path lists, single home today.

## Candidate smells

- Duplicated config-location fact across `read_source_roots` and `read_instruction_files` — will drift once `.uncoded.toml` adds a second home.
- `_find_project_root` and `find_pyproject_toml` hardcode `pyproject.toml` in the walk, the return, and the error text — the Python-only assumption the language-neutral goal must remove.
- `_replace_or_append`'s single-pair model — will not carry two sections or section removal.

alimanfoo:
# Session Scope

The accepted scope for this work is the Maximal option: the full coherent feature plus dogfooding it on this repo.

## Scope items

- **C1 — Opt-in doc-roots config.** uncoded reads an optional `doc-roots` setting: a list of directories (each walked for `*.md`) and/or individual `.md` files. Absent or empty means no doc indexing, so Python-only repos are unaffected.
- **C2 — Language-neutral config home.** Config is discovered by walking up for `pyproject.toml` or `.uncoded.toml`; `pyproject.toml` wins if both exist; config is never read from inside `.uncoded/`. The project root derives from wherever the config home is found, not from `pyproject.toml` specifically.
- **C3 — Single home for the config-location fact.** Config is read through one home-discovery and section-access path, so adding `doc-roots` and the second config file does not leave the config location copied across readers.
- **C4 — docs.yaml artefact.** `uncoded sync` produces `.uncoded/docs.yaml`: a filesystem tree of the configured doc-roots, with each `.md` file nesting its ATX heading hierarchy by level. It is independent of `namespace.yaml`, follows the same pure-key-hierarchy, no-values style, and carries its own header. ATX headings only; a `#` inside a fenced code block is not a heading.
- **C5 — Two instruction sections.** The injected instruction content provides a code-navigation section and a docs-navigation section, each present only when its root-type is configured and removed when its root-type is not. The docs section is short: orient via `docs.yaml`, then Read or grep to the literal heading. A docs-only repo gets no Python dispatch-rule noise.
- **C6 — Per-root-type sync and check independence, with cleanup.** `sync` and `check` handle each root-type's artefacts independently. Source roots present builds the code artefacts (`namespace.yaml`, stubs, code section); doc roots present builds the doc artefacts (`docs.yaml`, docs section). When a root-type is absent, its file artefacts are removed: no source roots removes `namespace.yaml` and stubs; no doc roots removes `docs.yaml`. At least one root-type is required, otherwise `sync` and `check` error with "nothing to index". `check` returns non-zero when any artefact is stale, including one that should be removed.
- **C7 — User-facing docs updated.** README and the hand-maintained prose in AGENTS.md describe doc indexing: the `doc-roots` setting, the `.uncoded.toml` home, and the `docs.yaml` artefact, so a maintainer can discover the feature.
- **M1 — Dogfood doc indexing in this repo.** Add `doc-roots` to uncoded's own config so the repo generates its own `docs.yaml` and docs-navigation section, exercising the feature on real docs.

Each scope item names an outcome; the implementation approach is decided in design.

alimanfoo:
# Design

## Config (`config.py`)

A frozen `Config` dataclass holds `project_root`, `source_roots`, `doc_roots`, and `instruction_files`. A single `read_config(start)` walks up for the config home, parses it, and reads every field in one pass. It returns `None` only when no config home is found; a home that exists but configures neither root-type yields a `Config` with empty roots. This collapses the previous `read_source_roots` and `read_instruction_files` and folds in the CLI's project-root discovery. `find_pyproject_toml` stays as-is — `refs.py` uses it to locate the Python project for its language-server exchange, which is a separate concern from where uncoded config lives.

`.uncoded.toml` uses top-level keys (`source-roots = [...]`, `doc-roots = [...]`); `pyproject.toml` keeps its `[tool.uncoded]` section. Discovery walks up for either file and prefers `pyproject.toml` when both sit in the same directory.

## Doc extraction (`docs_map.py`, new)

Mirrors `namespace_map.py`. `extract_headings` scans each Markdown file line by line, tracks fenced-code-block state, and collects ATX headings. `iter_doc_files` accepts a directory (walked for `*.md`) or an individual `.md` file. `build_docs_map` builds the nested dict; `DOCS_HEADER` is its own header comment.

## Shared rendering (`yaml_tree.py`, new)

The clean YAML dumper and a `render_yaml_tree(header, mapping)` helper move to one neutral module, imported by both `namespace_map` and `docs_map`. Neither artefact module depends on the other, and the pure-key-hierarchy style is single-sourced.

## Instruction sections (`instruction_files.py`)

The existing `<!-- uncoded:start -->` / `<!-- uncoded:end -->` pair stays as the code-navigation section, so existing instruction files need no migration. A new `<!-- uncoded:docs:start -->` / `<!-- uncoded:docs:end -->` pair carries the docs-navigation section. A generalised `_apply_section(text, start, end, body)` replaces, appends, or — when `body` is `None` — removes a section. `sync_instruction_file` applies both sections, passing the section body when its root-type is configured and `None` when it is not. The code body comes from `dispatch_rule.md`; a new short `docs_rule.md` resource is the docs body (orient via `docs.yaml`, headings are literal, Read or grep to the section).

## Orchestration (`cli._sync`)

Read the config once. If no home is found, print the "no config home" error and return 1. Validate that at least one root-type is configured, else print "nothing to index" (naming both keys) and return 1. Build each configured root-type's artefacts and remove the artefacts of any absent root-type. Sync both instruction sections (the body, or `None` to remove). Sync the skill. Sum the changes.

## Input contract

- ATX headings only: one to six `#` characters, a space, then text. Seven or more `#`, Setext underlines (`===`/`---`), and `#` lines inside fenced code blocks are not headings. Trailing closing `#` characters are stripped.
- A Markdown file with no headings still appears in `docs.yaml` as a null leaf, so an agent knows the file exists.
- Duplicate sibling headings get lossless ` (2)`, ` (3)` suffixes, incrementing past any real heading that already carries the suffix.
- A configured doc-root must exist and be a directory or a `.md` file, otherwise sync errors as it already does for an invalid source-root.
- Heading text with YAML-special characters is quoted by the shared dumper, as it already is for code symbols.

## Dogfooding

This repo adds `doc-roots` to its own config so it generates its own `docs.yaml` and docs-navigation section.

## Alternatives considered

No competing whole-design alternative emerged; the variation was in localized calls, each decided above:

- Config as a typed dataclass, over a thin dict return or keeping three separate readers.
- `.uncoded.toml` with top-level keys, over repeating `[tool.uncoded]` in the dedicated file (which would be copy-paste-able but carries a redundant wrapper).
- A shared renderer, over a standalone duplicate dumper in `docs_map`.
- A generalised section helper with `None`-means-remove and the legacy markers kept for code, over renaming both marker pairs with a one-time migration (which would churn every instruction file once) or two separate named functions.
- Lossless heading-suffix disambiguation, over first-wins or last-wins (both lose orientation entries silently).
- The at-least-one-root check at the sync operation, over a Config smart constructor, since `Config` is consumed only there.

alimanfoo:
# Plan

Nine tasks, run in order, one commit each. Each code task ends by regenerating the repo's own index so the committed `.uncoded/` stays current, keeps the 100% branch-coverage gate, and passes the type check and linter.

1. **Extract the shared YAML renderer.** Move the clean YAML dumper and the dump call out of `namespace_map` into a new `yaml_tree.py` as `render_yaml_tree(header, mapping)`, and repoint `namespace_map.render_map` to delegate. A behaviour-preserving refactor; the generated `namespace.yaml` is byte-identical.

2. **Introduce `Config` and `read_config`.** Add a frozen `Config` (project root, source roots, doc roots, instruction files) and a single `read_config` that walks up for `pyproject.toml`, parses `[tool.uncoded]` once, and reads all keys. Remove the two old readers and the CLI's project-root discovery. Replace the old "no section" error with a "nothing to index" guard in `_sync` that fires when neither root type is configured and names both keys. Existing configs behave exactly as before.

3. **Add `.uncoded.toml` as a second config home.** Extend discovery to find `.uncoded.toml` too: the nearest ancestor directory holding either file is the home, and `pyproject.toml` wins when both sit in one directory. Config is read as top-level keys in `.uncoded.toml`, versus `[tool.uncoded]` in `pyproject.toml`, and never from inside `.uncoded/`. Update the no-config-home error to name both files.

4. **Add the ATX heading parser.** A new `docs_map.py` gets `extract_headings`: one to six `#` then a space then a non-empty title, with trailing `#` stripped. Seven or more `#`, Setext underlines, and `#` lines inside fenced code blocks are not headings.

5. **Build and render the docs map.** Add `iter_doc_files` (a directory walked for `*.md`, or a single `.md` file), `build_docs_map` (heading hierarchy nested by level, duplicate siblings disambiguated with ` (2)`, ` (3)` …, a headingless file kept as a null leaf), and `render_docs_map` (its own header via the shared renderer).

6. **Split the instruction content into two sections.** Generalise the replace-or-append helper into one that also removes a section when its body is absent. Keep the existing markers as the code section, add a docs-section marker pair and a short new docs rule resource, and update the writer to apply both sections. Update the call site and the existing tests so the tree stays green.

7. **Wire doc artefacts and cleanup into sync.** In `_sync`, build each configured root type's artefacts and remove the artefacts of any absent root type. Validate each doc-root is a directory or a `.md` file. Apply both instruction sections by presence. Check mode covers the removals.

8. **Document doc indexing.** Update the README and the hand-maintained AGENTS.md prose to describe doc-roots, the `.uncoded.toml` home and precedence, and the `docs.yaml` artefact.

9. **Dogfood doc indexing in this repo.** Add doc-roots to this repo's own config covering its top-level Markdown, regenerate, and commit the generated `docs.yaml` and docs section.

alimanfoo:

Ready to merge — the diff delivers every agreed scope item with no gaps or coherence problems.

### Completeness against the agreed scope

- **Opt-in doc-roots config:** `read_config` reads `doc-roots` from `[tool.uncoded]` (pyproject) or top-level (`.uncoded.toml`). Absent or empty leaves existing behaviour unchanged.
- **Language-neutral config home:** `_find_config_file` walks up for either file, skips `.uncoded/` directories, pyproject wins in the same directory, nearer file wins across levels.
- **Single config-location home:** `read_config` does discovery, parse, and all three field reads in one pass. The old `read_source_roots`, `read_instruction_files`, and `_find_project_root` are gone.
- **docs.yaml artefact:** `docs_map.py` builds and renders the pure-key hierarchy with ATX-only scanning, fenced-block suppression, and its own header. Independent of namespace.yaml.
- **Two independent instruction sections:** `_apply_section` handles write/replace/remove for each section independently. Bare markers are the code section (backward compatible); the docs markers are the docs section. Each present only when its root type is configured, removed when not.
- **Per-root-type sync/check with cleanup:** `_sync` branches independently on source roots and doc roots. Absent source roots removes namespace.yaml and prunes the stubs tree. Absent doc roots removes docs.yaml. "At least one root type" error in place. Check mode counts removals.
- **User-facing docs:** README covers docs.yaml, doc-roots, the `.uncoded.toml` schema and precedence, and the updated "How agents use it" step. AGENTS.md adds the doc-map artefact and updates the sync comment.
- **Dogfood:** `doc-roots = ["README.md", "AGENTS.md"]` in this repo's config; docs.yaml generated and committed; docs section injected into AGENTS.md; `uncoded check` clean.

No issues. Marker non-collision holds: `<!-- uncoded:end -->` is not a substring of `<!-- uncoded:docs:end -->`, so the section finder always resolves correctly in a file carrying both. Backward compatibility holds: existing source-only repos get the same artefacts as before, and the docs.yaml removal is a no-op when the file is absent.

### Out of scope but noticed

1. `@dataclass(frozen=True)` with `list` fields prevents reassignment but not list mutation. No impact here; worth knowing if `Config` gains more consumers.
2. `render_docs_map(m)` names its parameter `m` while the parallel `render_map(namespace)` uses a descriptive name. Minor readability gap.

alimanfoo:
## Review — fresh-eyes read

**Recommendation:** Looks good — a few non-blocking things and small nits, nothing blocking.

My cold read of the diff matched the PR's stated intent closely. The doc-map feature, the per-root-type independence, and the `.uncoded.toml` config home all read the way they were meant to.

### Non-blocking

1. **Headings ending in a literal `#` get truncated.** `extract_headings` strips trailing `#` unconditionally. CommonMark only treats a trailing run of `#` as a closing sequence when a space precedes it. So a heading like `## Using C#` lands in docs.yaml as `Using C`, and an agent grepping for the real heading text won't find it. The tests pin the always-strip behaviour but never test a `#` that is part of the title.

2. **A `pyproject.toml` with no `[tool.uncoded]` shadows a sibling `.uncoded.toml`.** `_find_config_file` returns the first file found in a directory, trying `pyproject.toml` first. If that file has no uncoded section, sync errors with "nothing to index" even when a valid `.uncoded.toml` sits beside it. This bites a non-Python repo that keeps a `pyproject.toml` for any unrelated tool. The requirement says "pyproject.toml wins," but it may have meant "wins when it actually carries uncoded config."

3. **A doc-root resolving outside the project root crashes.** A configured doc-root is validated as a dir or `.md` file but then passed to `iter_doc_files`, which calls `relative_to(project_root)`. A root like `../shared-docs` passes validation yet raises an unhandled `ValueError` instead of a clean error. The source-root path appears to share this shape (likely pre-existing).

### Nits

1. **Empty `.uncoded/stubs/` directory left behind.** When source-roots is removed, the stub-pruning path deletes every `.pyi` but keeps the stubs root, leaving an empty directory in generated output.

2. **Dead defensive branch.** `sync_instruction_file`'s both-sections-None guard can't be reached from the CLI, which errors with "nothing to index" first. Fine as defence for direct callers/tests; noting it is not exercised by the real path.

3. **PR description prose uses `doc_roots`/`source_roots` (underscores)**, but the actual config keys are hyphenated. The README uses the hyphenated form correctly; only the PR body mismatches.

### Readability

1. **The "remove all stubs" path is non-obvious.** When source-roots is absent, `_sync` calls `build_stubs(files=[], source_root=project_root)` to prune the whole stubs tree. This works only because of an internal detail of `_write_stubs`. The inline comment points at it, but a reader still has to load `_write_stubs` to trust it — the one spot the cold read could not verify from the diff alone. A small named helper would keep the intent local to the call site.

alimanfoo:
## Revised config-home precedence

Supersedes the config-home precedence rule in the Session Scope (C2) and the Design above.

A fresh-eyes review surfaced a case the original rule got wrong. The original rule was "when both `pyproject.toml` and `.uncoded.toml` exist in the same directory, `pyproject.toml` wins." Taken literally, a non-Python repo that keeps a `pyproject.toml` for some unrelated tool — but configures uncoded in `.uncoded.toml` — would have its `.uncoded.toml` silently ignored and sync would report "nothing to index". That defeats the `.uncoded.toml` escape hatch for exactly the consumer it exists for.

The revised rule, walking up from the start directory and skipping the generated `.uncoded/` directory, is:

1. `pyproject.toml` wins only when it carries a `[tool.uncoded]` section.
2. A `pyproject.toml` with no `[tool.uncoded]` section no longer shadows a sibling `.uncoded.toml` — the `.uncoded.toml` is used.
3. A directory holding both a `[tool.uncoded]` `pyproject.toml` and a `.uncoded.toml` is a configuration error: uncoded must be configured in one file only.
4. Across different directories, the nearer config file still wins, with no error — ambiguity is only a same-directory condition.

A `pyproject.toml` that exists with no `[tool.uncoded]` section and no sibling `.uncoded.toml` still anchors the project and yields the existing "nothing to index" error, unchanged.

alimanfoo:
## Response to the reviews

Thanks to both reviewers. How each finding was handled:

**Completeness / coherence review** — found the agreed scope fully delivered with no gaps.

- Frozen dataclass with list fields — held for post-merge triage; no impact today.
- `render_docs_map` parameter named `m` — addressed in e1ec882 (renamed to `mapping`, matching the sibling renderer).

**Fresh-eyes review**

- Headings ending in a literal `#` were truncated (e.g. `C#`) — addressed in 464da7a. A trailing `#` run is now stripped only when it is a real closing sequence.
- A `pyproject.toml` with no `[tool.uncoded]` shadowed a sibling `.uncoded.toml` — this was the right call to revisit. The precedence rule is now: `pyproject.toml` wins only when it carries `[tool.uncoded]`; a bare `pyproject.toml` no longer shadows a sibling `.uncoded.toml`; and a directory holding both a `[tool.uncoded]` `pyproject.toml` and a `.uncoded.toml` is a configuration error. Addressed in 6161f30, with the README updated in c381312. See the "Revised config-home precedence" note above.
- A doc-root resolving outside the project root crashed — addressed in 10bc22a with a clear error before any path-relativisation runs.
- Empty `.uncoded/stubs/` directory left behind, and the remove-all-stubs path was opaque — both addressed in 23b1c99 with a named `remove_all_stubs` helper that also removes the now-empty directory.
- The unreachable both-sections-`None` guard in the instruction-file writer — left as intentional defence for direct callers; it is covered by a test.
- The PR description used underscored config keys — corrected in the description.

src/uncoded/cli.py
```diff
@@ -5,106 +5,159 @@
...
+                source_root=src_root,
+                output_dir=Path(".uncoded/stubs"),
+                project_root=project_root,
+                check=check,
+            )
+    else:
+        if remove_file(
+            Path(".uncoded/namespace.yaml"), project_root=project_root, check=check
+        ):
+            changes += 1
```
alimanfoo: Why not `changes += remove_file(...)` so the same pattern is used throughout to accumulate changes?

src/uncoded/cli.py
```diff
@@ -5,106 +5,159 @@
...
+            m for _root, files in roots_with_files for m in extract_modules(files)
+        ]
+        map_content = render_map(build_map(modules))
+        if sync_file(
+            Path(".uncoded/namespace.yaml"),
+            map_content,
+            project_root=project_root,
+            check=check,
+        ):
+            changes += 1
```
alimanfoo: Why not `changes += sync_file(...)`?

src/uncoded/cli.py
```diff
@@ -5,106 +5,159 @@
...
+        resolved_project_root = project_root.resolve()
+        for configured in config.doc_roots:
+            doc_root = (project_root / configured).resolve()
+            if not doc_root.is_relative_to(resolved_project_root):
+                print(
+                    f"Error: doc root {configured} is outside the project root. "
+                    "Check doc-roots in your uncoded config file.",
+                    file=sys.stderr,
+                )
+                return 1
```
alimanfoo, line 132: Why do we check this for `doc_root` but not for `src_root`?

And anyway, how could `doc_root` ever be outside the project root?

alimanfoo (commented):
Some questions inline...

src/uncoded/cli.py
```diff
@@ -114,7 +170,7 @@ def _sync(*, start: Path | None = None, check: bool = False) -> int:
     # ``project_root`` for the user-facing line, falling back to the
     # absolute resolved path when the file lives outside ``project_root``.
     seen_resolved: set[Path] = set()
-    for path in read_instruction_files(project_root):
+    for path in config.instruction_files:
         resolved = (project_root / path).resolve()
```
alimanfoo: Now here it looks like we allow instructions files outside the project root. But that's not now consistent with src and doc roots.

alimanfoo (commented):
One comment inline...

src/uncoded/docs_map.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+def build_docs_map(files: Iterable[tuple[str, Path]]) -> dict:
+    """Build a nested dict from ``(text, rel_path)`` doc-file pairs.
+
+    Top-level keys are directory names (trailing "/") and file names.
+    Headings nest under the nearest preceding heading of a smaller level;
+    level skips attach to the nearest shallower heading. Duplicate sibling
+    headings are disambiguated with " (2)", " (3)", … A file with no
+    headings maps to null.
```
alimanfoo, line 120: How is an agent supposed to know that "foo bar (2)" means the second occurrence of header "foo bar"?

src/uncoded/docs_rule.md
```diff
@@ -0,0 +1,6 @@
+## How to read documentation in this codebase
+
+Load `.uncoded/docs.yaml` to orient: it lists every Markdown file and its
+heading hierarchy. Headings are literal text in the source file — use `Read`
+or `grep` to navigate to a section. There are no `uncoded body`, `uncoded
+refs`, or stub files for Markdown.
```
alimanfoo: "There are no `uncoded body`, `uncoded refs`, or stub files for Markdown."

This would read oddly if in a docs-only repo.

amiles-eit:
## Follow-up: resolutions to the later review comments

Addressing the inline review questions and comments raised after the PR was marked ready:

- **Change accounting (`if writer(): changes += 1` vs `changes += writer()`)** — unified to `changes += <writer>(...)` across all the bool-returning writers in `_sync`, matching the int-returning ones (101fdf1).
- **Why guard `doc_root` but not `src_root`, and how could a root be outside the project?** — a root can resolve outside via a `..` path, an absolute path, or a symlink. Source roots had the same latent crash, so the guard is now shared: a configured **source or doc root** outside the project root is rejected with a clear error (ffca226).
- **Instruction files appeared to allow paths outside the project root** — now rejected too, so every configured path follows one rule: it must be within the project root (90da62f). With that rule active at the boundary, the leaf writers' advertised absolute-path bypass was removed — the docstring contract and its test dropped from `sync_file`, `remove_file`, and `sync_instruction_file` (d8aaac6).
- **How does an agent know `foo bar (2)` means the second occurrence?** — `docs.yaml`'s header now documents the convention: a ` (N)` suffix marks the Nth occurrence of a repeated heading, and you grep the base text and take the Nth match (dcaa2a0).
- **The docs navigation section read oddly in a docs-only repo** — rephrased to stand alone, dropping the contrast with the code-only tools (4fc9cc8). The "outline only, no body/refs/stubs" point stays in the hand-maintained README/AGENTS prose where it fits.

PR #172, opened 2026-06-18
Fix two sync write defects in docs-only and formatter setups

### LINKED ISSUE #170: sync writes the coherence-review skill even with no source-roots
`_sync` calls `sync_skill` unconditionally (`src/uncoded/cli.py:193`), so
`uvx uncoded sync` always writes `.claude/skills/coherence-review/SKILL.md` and
the `.agents/` twin — even in a docs-only repo configured with only `doc-roots`.

But the skill is a **code** artefact: its own prerequisites require
`.uncoded/namespace.yaml` and `.uncoded/stubs/`, which `_sync` *removes* when
`source-roots` is unset (`cli.py:111-119`). So in a docs-only repo the skill is
dead on arrival — it can never run, yet it installs itself into the agent's
skill list (it auto-loads in Claude Code).

## Fix

Treat the skill like the other code artefacts: build it when `source-roots` is
configured, remove it otherwise. Mirror the existing namespace/stubs
build-or-remove rule:

```python
# cli.py
changes += sync_skill(project_root=project_root, build=bool(config.source_roots), check=check)
```

sync_skill writes SKILL_OUTPUTS when build is true and remove_files them otherwise; legacy paths stay always-removed. This removes a special case rather
than adding a config knob — the skill simply joins the other code artefacts under the same rule.

Tests in tests/test_skill.py and tests/test_cli.py need the build
parameter threaded through (e.g. a docs-only _sync run should leave no skill
files, and should remove a pre-existing one).

### LINKED ISSUE #171: Instruction-file injection re-asserts exact bytes, conflicting with markdown formatters
The navigation section uncoded injects into CLAUDE.md/AGENTS.md is **necessary** —
it's how an agent learns the index exists and how to read it. Disabling it
(`instruction-files = []`) makes the index write-only, so opting out isn't a real
solution; the section has to be present.

The problem is that when the section already exists, `_apply_section`
(`src/uncoded/instruction_files.py:53-56`) overwrites the bytes between the
markers with uncoded's exact `body` on every `sync`. A repo that formats its
markdown reflows that block — e.g. prettier with `proseWrap: always` rewraps the
prose to a different width and inserts a blank line after the opening
`<!-- uncoded:...:start -->` comment. On the next run `uncoded check`'s
byte-identical comparison fails, and `sync` rewrites the block back to its own
form. Neither tool accepts the other's output, so a pre-commit setup running both
never converges.

## Reproduce

```toml
# .uncoded.toml
doc-roots = ["docs"]
```

```json
// .prettierrc
{ "proseWrap": "always", "printWidth": 80 }
```

1. `uvx uncoded sync`        # injects the docs section into AGENTS.md
2. `prettier --write AGENTS.md`   # reflows the injected block
3. `uvx uncoded check`       # fails: "Would update AGENTS.md"
4. `uvx uncoded sync`        # rewrites it back
5. `prettier --check AGENTS.md`   # fails again — loop never settles

## Fix

Let uncoded keep the section present while tolerating a formatter's reflow:

- **Inject-if-missing (preferred):** write the section when the markers are
  absent, then leave it untouched. `check` fails only if the markers go missing,
  so the formatter is free to own the wrapping. The guidance still lands (agents
  get it) and the fight ends. Cost: uncoded can no longer refresh the boilerplate
  wording across versions — recoverable later with a version marker in the
  comment if that ever matters.
- **Whitespace-tolerant compare:** only rewrite when the body differs after
  normalizing wrapping/whitespace. Avoids even the first-injection hiccup, but
  "normalize markdown" is fuzzier and more code.

Inject-if-missing also keeps the model clean: uncoded plants the pointer once and
the repo owns the bytes thereafter — the same "seed, not a managed mirror"
relationship that suits a near-static boilerplate block.

alimanfoo:
This PR fixes two defects in what `uncoded sync` writes to a repo, both surfaced by the docs-only support added in #164. It addresses #170 and #171.

## Session type

Bug fix.

## Bug 1 — skill written even with no source-roots (#170)

**Expected behaviour.** The coherence-review skill is a code artefact. It should be written only when `source-roots` is configured, and removed otherwise — mirroring `namespace.yaml` and stubs, whose absence (when `source-roots` is unset) leaves the skill's diagnostic sweeps nothing to run against.

**Observed behaviour.** `_sync` calls `sync_skill` unconditionally, so a docs-only repo gets `.claude/skills/coherence-review/SKILL.md` and the `.agents/` twin written — a skill that can never run but auto-loads into the agent's skill list.

**Affected consumers.** A maintainer of a docs-only repo (for example a non-Python repo configured with only `doc-roots`) running `uvx uncoded sync`. They get a dead, auto-loading skill polluting their agent's skill list.

## Bug 2 — instruction-file injection fights markdown formatters (#171)

**Expected behaviour.** The navigation section stays present, and:

- Within a given uncoded version, a repo's markdown formatter is free to reflow the section. `sync` and `check` tolerate the reflow and converge — they do not rewrite it back.
- When uncoded is upgraded to a version whose section wording differs, the section is refreshed to the new wording.

**Observed behaviour.** `_apply_section` overwrites the bytes between the markers with uncoded's exact body on every `sync`. A formatter (for example prettier with `proseWrap: always`) reflows the block; `check` then fails and `sync` rewrites it — the pre-commit loop never settles.

**Affected consumers.** A maintainer who runs a markdown formatter on `CLAUDE.md` or `AGENTS.md` alongside uncoded in pre-commit. Cost: pre-commit never converges; they must disable one tool or the section.

## System non-goals

None new.

Closes #170
Closes #171


alimanfoo:
## Code Analysis

### Architecture the work touches

`_sync` (`cli.py`) is the orchestrator. It builds each artefact under a per-root-type build-or-remove rule: an artefact tied to a root type is written when that root type is configured, and removed when it isn't.

- Code artefacts gated on `source_roots`: `namespace.yaml` and stubs are built in the `if config.source_roots:` block and removed (`remove_file` / `remove_all_stubs`) in its `else`.
- Doc artefact gated on `doc_roots`: `docs.yaml` is built/removed the same way.
- Instruction sections: `code_section = SECTION_CODE if config.source_roots else None` and `docs_section = SECTION_DOCS if config.doc_roots else None` — each present only when its root type is configured.

This rule is a convention, enforced only by `_sync` applying it by hand for each artefact. Nothing checks that a new artefact is gated on its root type.

### Bug 1 — root cause (#170)

The skill is the one code artefact that breaks the convention. The final line of `_sync` is `changes += sync_skill(project_root=project_root, check=check)` — unconditional, outside any `source_roots` branch. `sync_skill` (`skill.py`) has no build/remove switch: it always writes `SKILL_OUTPUTS` and always removes `LEGACY_SKILL_OUTPUTS`.

So in a docs-only repo (`source_roots` empty), `_sync` removes `namespace.yaml` and stubs but still writes `.claude/skills/coherence-review/SKILL.md` and the `.agents/` twin. The skill's diagnostic sweeps need exactly those removed artefacts, so it is dead on arrival yet auto-loads into the agent's skill list.

The root cause is the missing gate, not anything inside `sync_skill` — the skill needs to join the other code artefacts under the build-or-remove rule (a `build` parameter threaded from `_sync`, matching `bool(config.source_roots)`).

Test gap: `TestSyncDocRoots` (`test_cli.py`) asserts a docs-only run writes no `namespace.yaml` and no stubs, but says nothing about the skill — so the current dead-skill behaviour is unpinned.

### Bug 2 — root cause (#171)

The instruction-file injection is a managed mirror, not a one-time plant. The chain:

- `SECTION_CODE` / `SECTION_DOCS` (`instruction_files.py`) are exact-byte constants — marker + body + marker, with no version stamp in the markers (`<!-- uncoded:start -->` and so on).
- `sync_instruction_file` reads the file, re-applies the section via `_apply_section` (pure string surgery: find the markers, replace the bytes between them with the canonical constant), and passes the recomputed full text to `sync_file`.
- `sync_file` (`sync.py`) writes whenever `target.read_text() != content` — a byte comparison.

So any reflow of the bytes between the markers differs from the canonical constant, and `sync` rewrites it (`check` returns 1). A markdown formatter reflowing the block and uncoded re-asserting it never converge.

The deeper cause is that uncoded has no version awareness here: a byte diff from a formatter's reflow and a byte diff from stale wording (an old uncoded version) look identical, so the code can only do the one thing — reassert exact bytes. The requirement (tolerate reflow within a version, refresh on upgrade) needs the section to carry its version and the compare to key on version, not bytes. The version is available at runtime as `uncoded.__version__` (`__init__.py`, derived from git tags via hatch-vcs), but `instruction_files.py` neither imports nor uses it today.

### Note (factual, not a proposal)

The "build-or-remove every artefact per its root type" convention has nothing checking it — Bug 1 is what its absence let through. Whether to add such a check is Scope's call.

alimanfoo:
## Session Scope (Coherent)

1. The coherence-review skill follows the same per-root-type build-or-remove rule as the other code artefacts: written when `source-roots` is configured, removed when it isn't — including removing a pre-existing skill in a docs-only repo. (Fixes #170.)
2. A regression test pins the skill's docs-only behaviour: a docs-only sync writes no skill files and removes a pre-existing one; check mode reports the removal without writing.
3. Instruction-section sync converges with a markdown formatter within an uncoded version: once the section is present, a formatter's reflow of the section body no longer triggers a rewrite by `sync` or a failure by `check`. (Fixes #171.)
4. The instruction section refreshes when uncoded is upgraded to a version whose section wording differs.
5. Regression tests pin Bug 2 behaviour: within a version a reflowed section survives sync/check (convergence); on a version change the section is refreshed; markers-absent still injects.

Only the Coherent Scope applies: there is no meaningful narrower option (regression tests are part of a bug fix, and a no-refresh fix for #171 sits below the agreed requirement), and no real wider anticipation.

A check enforcing the "build-or-remove every code artefact per its root type" convention was considered and rejected. The structural fix for #170 brings the skill call inside the same `source-roots` branch as the other code artefacts, making the convention visible in the code and removing the special case. That beats a brittle custom check with no off-the-shelf tool.

alimanfoo:
## Design

### Bug 1 (#170) — gate the skill via a `build` parameter

Add a `build: bool` keyword to `sync_skill`: `True` writes `SKILL_OUTPUTS`, `False` removes them, and `LEGACY_SKILL_OUTPUTS` is removed unconditionally either way. `_sync` calls it once with `build=bool(config.source_roots)`, grouped with the other code artefacts. All skill logic (write, remove, legacy cleanup) stays in `skill.py`.

### Bug 2 (#171) — content fingerprint in the opening marker

The opening marker carries a short fingerprint of uncoded's canonical section body: `<!-- uncoded:start sha256=ab12cd34… -->` (code) and the docs equivalent. The body is static per release, so the fingerprint is computed once at module load and the section constants stay constants. `_apply_section` locates an existing section by the stable prefix (`<!-- uncoded:start`) and acts in three cases:

- absent → inject the canonical section (with current fingerprint);
- fingerprint matches → leave the file untouched, so a markdown formatter's reflow survives and `sync`/`check` converge;
- fingerprint differs → refresh the section to the canonical wording.

Refresh fires exactly when the wording changes — never on a version bump that leaves the wording identical — so there is no formatter re-fight. No `__version__` import is needed; the content is its own version. Repos carrying the old plain marker refresh once on the first sync after upgrade, then stay stable.

### Alternatives considered

- **Version-string stamp.** Human-readable marker, but refreshes on every release even with unchanged wording, and churns on hatch-vcs dev versions.
- **Whitespace-normalised compare (no marker change).** No stamp, but "what a formatter does" is hard to fully anticipate; gaps reintroduce the loop.

alimanfoo:
## Plan

**Task 1 — Gate the coherence-review skill on source-roots (#170).** Add a `build: bool` keyword to `sync_skill` (write `SKILL_OUTPUTS` when True, remove them when False; remove `LEGACY_SKILL_OUTPUTS` unconditionally). `_sync` calls it with `build=bool(config.source_roots)`, grouped with the other code artefacts. Update existing skill-test callers to pass `build`. Regression tests: a docs-only sync writes no skill files and removes a pre-existing one; check mode reports the removal without writing.

**Task 2 — Make instruction-section injection fingerprint-aware (#171).** Embed a short sha256 fingerprint of each canonical section body in its opening marker; compute at module load. `_apply_section` locates a section by the stable opening prefix and: injects when absent, leaves the file untouched when the fingerprint matches (formatter reflow survives), refreshes when it differs. Update existing instruction-file tests to defend the behaviour, not the exact marker string. Regression tests: reflow survives sync/check within a version; a wording change refreshes; markers-absent injects; an old plain marker refreshes once then stays stable.

The two tasks are independent (`skill.py`/`cli.py` vs `instruction_files.py`), with no tidy-first precursors. Bug 1 first.

alimanfoo:

Recommend merge. The diff delivers both bug fixes in full, all regression tests are present, and the change reads coherently as a whole.

**Lens 1 — Completeness against requirements and scope**

All five scope items are delivered:

1. The skill follows the build-or-remove rule: `sync_skill` gains `build: bool`; `_sync` passes `build=bool(config.source_roots)` from the code-artefact region. Docs-only repos no longer receive a dead skill. ✓
2. Docs-only regression tests: `test_doc_only_does_not_write_skill`, `test_doc_only_removes_preexisting_skill`, and `test_check_returns_one_when_skill_should_be_removed` cover all three required cases (no write, removal, check-mode reporting). ✓
3. Formatter convergence within a version: `_apply_section` returns the file untouched when the opening marker fingerprint matches, so `sync_file`'s byte-compare sees no change and check passes. ✓
4. Refresh on wording change: the fingerprint-differs path replaces the whole section with the canonical body and updated marker. ✓
5. Regression tests for Bug 2: all four required cases are covered — `test_reflowed_body_survives_sync`, `test_reflowed_body_passes_check`, `test_different_fingerprint_refreshes_section`, `test_plain_marker_refreshes_once_then_stable`. The markers-absent inject case was already covered by pre-existing tests, and the inject path in `_apply_section` is unchanged; those tests still hold. ✓

One criterion applied mid-session and present in the full diff: `test_source_root_removal_cleans_code_artefacts` now asserts that skill files exist after the first sync and are absent after the second (source-roots dropped), consistent with the test's own name and the skill's new status as a code artefact. ✓

**Lens 2 — Coherence across the whole diff**

No gaps appear when reading the complete change together.

The `build` parameter is a required keyword-only argument, so all call sites were forced to update. The `test_skill.py` update to `build=True` throughout is complete and consistent.

The three existing `test_instruction_files.py` fixtures that previously used `MARKER_START` to simulate "old content" now use the literal plain marker `<!-- uncoded:start -->` as intended — the fixtures simulate pre-upgrade state, not current state.

The `test_cli.py` symlink-dedup test correctly imports `MARKER_START` rather than hard-coding the old string literal. AGENTS.md and CLAUDE.md in the repo itself carry the new fingerprinted markers, dogfooding the upgrade path.

The constant ordering in `instruction_files.py` (end-markers and bodies before fingerprinted start-markers) makes the dependency chain readable without forward references.

**Out of scope but noticed:** nothing.

alimanfoo:
## Review — fresh-eyes read

**Recommendation:** Looks good — no blockers; a few non-blocking items and readability notes worth a look.

A cold read matched the PR's stated intent on both fixes (skill gated on source-roots; markers fingerprinted to tolerate formatter reflow within a version and refresh on upgrade). Convergence holds in all normal paths and the committed fingerprints are consistent with the code, so CI's `uncoded check` should pass. The skill `build` flag is clean and complete — all callers updated, check-mode aggregation correct, ordering change cosmetic.

### Non-blocking

1. **The fingerprint is a version stamp, not a body integrity check — `sync`/`check` no longer heal body drift within a version.** In `_apply_section` (`src/uncoded/instruction_files.py`), a section whose opening-marker line equals the current marker is returned untouched no matter what the body now contains. The hash is computed from the *canonical* body at module load and never compared against the on-disk body. Before this PR, `sync` always rewrote the body, so any drift (a bad merge, an accidental deletion, half the section lost) self-healed and `check` flagged it. Now only formatter reflow was the target, but the same mechanism silently tolerates *any* body change. Worth confirming the team accepts losing that self-healing guarantee — it is a real behaviour change beyond the stated formatter case.

2. **Section is located by `text.find` on a prefix substring, which can silently delete content.** `marker_prefix = "<!-- uncoded:start"` is matched anywhere in the file, not just at the start of a line. If an instruction file mentions uncoded's own marker syntax in prose *before* the real section, `find` anchors on that prose; the replace then spans from there to the next end marker and deletes the text in between. This is a pre-existing class (the old full-marker `find` had it too), but switching to a prefix widens it. Anchoring the search to the start of a line would close it.

### Nits

1. **CRLF line endings leave a stray `\r` and force one extra rewrite.** `after = text[e + len(end):].lstrip("\n")` strips `\n` but not `\r`, and `existing_opening` keeps the `\r` before the newline, so a CRLF file never matches the fingerprint on first sync. It converges after one rewrite but leaves mixed line endings around the section. `lstrip("\r\n")` and tolerating `\r` in the marker-line compare would fix it. Confirm whether CRLF inputs are in scope at all.

2. **Duplicate sections are only half-handled.** `text.find` operates on the first occurrence only. If a file somehow holds two start-prefixed sections, only the first is updated or removed; the second is left orphaned permanently. Requires malformed input, so low priority.

### Out of scope but noticed

1. **A file with the end marker before the start marker never settles.** `section_found` is False, so each `sync` appends a fresh canonical section without removing the misplaced one — the loop never converges. Pre-existing, not introduced here; needs manual repair.

### Readability

1. **"Fingerprint" oversells the mechanism.** The module comment frames the hash as detecting whether the body "may have been reformatted." In practice the on-disk body is never inspected; the hash is only a version stamp on the marker line. Naming it a version/wording stamp would set a reader's expectations correctly — see non-blocking 1.
2. **Why the skill is gated on source-roots isn't visible in the code.** The reasoning (the skill's sweeps have nothing to run against without source) lives only in the PR description. Reading `cli._sync` alone, `build=bool(config.source_roots)` looks arbitrary. A one-line comment at the call site would carry the intent forward.
3. **`marker_prefix = "<!-- " + start.split(" ")[1]` reads as clever-by-reconstruction.** A reader must mentally run the split to see it yields `<!-- uncoded:start`, and it assumes the marker always has a space-delimited second token. The adjacent comment helps; a more direct derivation would help more.

alimanfoo:
## Response to the reviews

The completeness-and-coherence review raised no findings and recommended merge — no action needed there.

The fresh-eyes review:

**Non-blocking 1 — fingerprint is a version stamp, not a body integrity check; drift within a version no longer self-heals.** Reject. This is the intended behaviour: the agreed requirement is that within a given uncoded version the repo owns the bytes between the markers, so a formatter's reflow survives. Tolerating body edits within a version is the deliberate trade ("seed, not a managed mirror"). The related framing concern is addressed under Readability 1 below.

**Non-blocking 2 — prefix match not anchored to a line start could delete content.** Accepted. Fixed in 58b14b9: the opening marker is now located with a line-anchored match, so marker-like text in prose can no longer be mistaken for a section opener. A regression test covers it.

**Readability 1 — "fingerprint" comment oversells the mechanism.** Accepted. Fixed in 58b14b9: the module comment now describes a stamp comparison on the marker line (the on-disk body is never inspected), refreshing only when the wording changes.

**Readability 2 — why the skill is gated on source-roots isn't visible.** Accepted. Fixed in 766f722: a comment at the call site explains the skill needs the code index to run.

**Readability 3 — prefix derivation reads as clever-by-reconstruction.** Accepted. Fixed in 58b14b9: the stable prefix now comes from explicit `MARKER_START_PREFIX` / `MARKER_DOCS_START_PREFIX` constants.

Held for post-merge triage (pre-existing, none introduced by this PR): Nit 1 (CRLF line endings force an extra rewrite), Nit 2 (only the first of duplicate sections is handled), and the out-of-scope note (an end-marker before the start-marker never settles).

PR #177, opened 2026-06-19
Cleanup follow-ons from #165 and #172

### LINKED ISSUE #163: Generated AGENTS.md/CLAUDE.md dispatch-rule prose uses "is the rule firing" three times — hard to parse
The agent-navigation prose generated into `AGENTS.md` / `CLAUDE.md` by `uncoded sync` (inside the `<!-- uncoded:start -->...<!-- uncoded:end -->` block, in the "dispatch rule" section) uses the phrase **"is the rule firing"** three times in succession:

> `grep -rn 'def resolve_body'` to read a function's body **is the rule firing** — that is `uncoded body`. `grep -rn 'function_name'` to check whether something has callers before a refactor **is the rule firing** — that is `uncoded refs`. `grep` then `Edit` to delete dead code **is the rule firing** — that is `uncoded refs` to confirm dead, then `Edit`.

Read literally, "the rule firing" lands as a noun phrase; the intended meaning ("this grep usage is the dispatch rule's case — use the indexed tool instead") needs unpacking. This is the file every agent loads at session start, so the readability cost is paid often.

A small rewording around the three instances would make the section easier to skim on first read.

### LINKED ISSUE #167: Injected dispatch-rule examples reference uncoded's own symbols
Noticed during #165, which worked on the injected instruction-file sections.

`src/uncoded/dispatch_rule.md` teaches the dispatch rule with examples drawn from uncoded's own internals — for example `grep -rn 'def resolve_body'` and `function_name`. That text is injected verbatim into every consumer repo's `CLAUDE.md`/`AGENTS.md`, where `resolve_body` doesn't exist. An agent reading it there may hunt for a symbol that isn't in their codebase, or read generic guidance as if it were repo-specific.

Suggested direction: replace the self-referential examples with neutral placeholders (`<name_path>`, `<relative_path>`) or a plausible generic stand-in. Related to #163, which also concerns this file's injected prose.

### LINKED ISSUE #168: Nothing documents that the repo dogfoods uncoded via the pre-commit hook
Noticed during #165.

The repo runs uncoded on itself, but nothing tells a contributor so. The pre-commit hook `uv run uncoded sync` (in `.pre-commit-config.yaml`) regenerates `.uncoded/namespace.yaml`, the stub files, and now `.uncoded/docs.yaml` from local source on every commit. A contributor — human or agent — who doesn't know this reads the regenerated files as surprise changes, or sees the commit hook modify files and has to work out why before re-staging.

Suggested direction: a short note in `AGENTS.md` or a `CONTRIBUTING` file naming the dogfooding and what the hook regenerates, so the behaviour is expected rather than discovered.

### LINKED ISSUE #169: iter_doc_files aborts the whole sync on an unreadable or non-UTF-8 file
Noticed during #165, which added the Markdown doc index.

`iter_doc_files` in `src/uncoded/docs_map.py` reads each file with `read_text(encoding="utf-8")` and no error handling, so a `.md` file that isn't valid UTF-8 — or that can't be read because of a permission error — raises an exception that propagates up and aborts the whole `sync` run. No index is built, rather than just that one file being skipped.

Its sibling reader for source, `iter_source_files` in `src/uncoded/extract.py`, already handles the analogous case: it catches the parse error, prints a one-line `warning: skipping <file>: ...` to stderr, and continues.

Suggested direction: give `iter_doc_files` the same treatment — catch `UnicodeDecodeError` and `OSError`, emit a skip warning, and continue — so one bad file degrades gracefully instead of taking down the whole index.

### LINKED ISSUE #173: _apply_section can corrupt or fail to converge on some instruction-file inputs
Noticed during #172. `_apply_section` (`src/uncoded/instruction_files.py`) is meant to keep a delimited section in sync convergently — running `sync` twice should equal running it once. Several inputs break that, and one can delete content.

1. End marker not line-anchored. #172 anchored the opening-marker search to the start of a line, but `text.find(end)` still matches `<!-- uncoded:end -->` anywhere. Prose containing the end marker on its own line is treated as the section boundary, so a replace can delete the text between it and the real markers. This is the mirror of the opening-side fix.
2. CRLF line endings. The marker-line compare keeps a trailing `\r`, and `lstrip("\n")` does not strip `\r`, so a CRLF file never matches the stamp on the first sync and ends up with mixed line endings. Convergence takes an extra rewrite.
3. Duplicate sections. Only the first prefix match is handled; a second start-prefixed section is left orphaned forever.
4. End marker before start marker. The section is treated as not found, so each sync appends a fresh section without removing the misplaced one — the file never settles.

Suggested direction: handle these in `_apply_section` (line-anchor the end marker as the start marker now is; tolerate CRLF; decide a policy for duplicate or misordered markers). A property-based test asserting "sync twice equals sync once" over arbitrary instruction-file contents would guard this whole class. Also add a unit test that the marker stamp really derives from the section body — today only CI's `uncoded check` on committed files would catch a broken derivation.

### LINKED ISSUE #174: Skill removal path and docs-section fingerprint lack direct unit tests
Noticed during #172. Two code paths added or touched by that PR are only covered indirectly.

`sync_skill`'s `build=False` removal path (`src/uncoded/skill.py`) is exercised only through the CLI integration tests in `tests/test_cli.py` (`TestSyncDocRoots`); `tests/test_skill.py` covers only `build=True`.

The fingerprint scenarios in `TestSyncInstructionFileFingerprint` cover the code section only. The docs section runs through the same shared `_apply_section`, so the behaviour is covered, but no test drives `sync_instruction_file` with `docs_section` under the reflow / different-stamp / plain-marker cases.

Suggested direction: add `build=False` unit tests in `test_skill.py` (removes existing files; returns False when already absent; check mode reports without removing) and parallel docs-section fingerprint tests. Low priority — the behaviour is already covered indirectly.

### LINKED ISSUE #175: Build-or-remove convention in _sync uses two different shapes
Noticed during #172. In `_sync` (`src/uncoded/cli.py`), every code artefact follows the same per-root-type rule — build it when `source-roots` is configured, remove it otherwise — but in two different shapes. Namespace map and stubs use an `if config.source_roots: … else: …` block; the skill uses `sync_skill(build=bool(config.source_roots))`. Both are correct, but a future author adding a new code artefact sees two models with no stated preference.

Suggested direction: name the convention in a short comment above the code-artefact region, or unify the two shapes. An enforcing check was considered and deliberately rejected during the #172 work, so this is about making the pattern clear, not enforcing it.

### LINKED ISSUE #176: Command guidance points contributors at uvx, which runs the published package not local changes
Noticed during #172, building on #155. The hand-written Commands section in `CLAUDE.md`/`AGENTS.md` tells readers to run uncoded commands via `uvx uncoded …`. #155 chose `uvx` so the examples work in consumer repos where uncoded is not a dependency — correct for consumers. But `uvx` runs the published package, so a contributor working on uncoded itself who follows that guidance does not exercise their working-tree changes.

During #172 this caused a sync mismatch: a local sync left the repo's own `AGENTS.md`/`CLAUDE.md` markers stale until the pre-commit hook — which uses `uv run uncoded sync` — regenerated them. #155 itself suggested documenting both forms with a note on when each applies; that note was never added.

Suggested direction: add the when-each-applies note — `uvx uncoded …` for consuming uncoded, and `uv run uncoded …` against the local editable install for developing uncoded (as the pre-commit hook already does). Check the README dev-setup section too.

alimanfoo:
Cleanup of follow-on issues from the recently merged #165 (Markdown doc index) and #172 (sync write-defect fixes). Maintenance session: coherence, robustness, test coverage, and honest docs of already-working code, plus one genuine bug fix.

## Requirements analysis

Session Type: maintenance. The unifying goal is to leave the post-#165/#172 codebase coherent, robust, tested, and honestly documented. Each goal below is stated as a checkable property.

### GH167 — injected examples reference uncoded's own symbols

The dispatch-rule text injected into a consumer's `CLAUDE.md`/`AGENTS.md` must contain no example naming a symbol that exists only in uncoded (`resolve_body`, `function_name`). `src/uncoded/dispatch_rule.md` examples use neutral placeholders or generic stand-ins.

### GH168 — dogfooding is undocumented

A contributor can learn from the repo's own docs that uncoded runs on itself via the pre-commit hook, and what `uncoded sync` regenerates.

### GH169 — `iter_doc_files` aborts the whole sync on one bad file

An unreadable or non-UTF-8 `.md` file is skipped with a one-line stderr warning and the sync continues, matching the skip-and-warn behaviour of its source-side sibling. `iter_doc_files` catches `UnicodeDecodeError` and `OSError`.

### GH173 — `_apply_section` can corrupt content or fail to converge (bug)

Expected (source: the function's own docstring — running sync twice equals running it once — and it must never delete user prose): the section update is convergent over arbitrary instruction-file contents and never deletes text outside the markers.

Observed (claim verified in analysis): four inputs break this — an unanchored end-marker search can delete prose containing the end marker; CRLF files never match the stamp and mix line endings; a duplicate section is orphaned forever; an end-before-start ordering appends endlessly.

Affected consumers: any repo whose instruction files contain marker-like prose, use CRLF, or have malformed or duplicate markers — the index write corrupts content or never settles.

### GH174 — skill-removal and docs-fingerprint paths lack direct unit tests

`sync_skill`'s `build=False` removal path and the docs-section fingerprint cases have direct unit tests, not only indirect CLI coverage.

### GH175 — build-or-remove uses two shapes in `_sync`

The per-root-type build-or-remove pattern for code artefacts reads as one model, so a future author adding a code artefact has one clear precedent. An enforcing check was deliberately rejected in #172 and stays rejected.

### GH176 — command guidance points contributors at `uvx`

The command guidance distinguishes `uvx uncoded …` (consuming uncoded) from `uv run uncoded …` (developing uncoded against local changes, as the hook does). Covers the hand-written Commands section in `CLAUDE.md`/`AGENTS.md` and the README dev-setup section.

### GH163 — repeated "is the rule firing" prose (promoted candidate)

The same injected dispatch-rule prose says "is the rule firing" three times in one paragraph and is hard to parse. Folded in because GH167 already opens `dispatch_rule.md` for editing.

### Source-read robustness (promoted candidate)

`iter_source_files` (`extract.py`) guards only `ast.parse`, not its `read_text`, so a non-UTF-8 or unreadable `.py` file aborts sync the same way GH169 describes for docs. Extend the graceful skip to source reads too.

### Preserved behaviour

- Index artefact content and format are unchanged for well-formed inputs.
- `_sync`'s per-root-type build/remove semantics stay identical.
- The dispatch rule's meaning is unchanged — only the example symbols change.

Closes #167
Closes #168
Closes #169
Closes #173
Closes #174
Closes #175
Closes #176
Closes #163


alimanfoo:
# Code Analysis

A structural read of the surfaces these nine goals touch, with citations.

## Architecture touched

Three concerns meet in `cli.py::_sync`, the one orchestrator: code-index artefacts (`namespace_map.py`, `stubs.py`, driven by `iter_source_files` in `extract.py`); doc-index artefacts (`docs_map.py`, driven by `iter_doc_files`); and injected guidance (`instruction_files.py` marker sections and `skill.py`). `_sync` applies one rule per root type — build the artefact when its root is configured, remove it otherwise. This rule holds only by convention; nothing checks it, and #172 deliberately rejected adding a check (GH175). Two reader siblings, `iter_source_files` and `iter_doc_files`, are meant to be parallel but diverge in error handling (GH169). The injected section body flows from two `.md` resource files (`dispatch_rule.md`, `docs_rule.md`) through a fingerprinted-marker mechanism.

## GH173 — `_apply_section` (`instruction_files.py:51`), the bug

The section locator is asymmetric. The start marker is line-anchored — `re.search(r"^" + re.escape(prefix), text, re.MULTILINE)` (line 67), a fix #172 landed. The end marker is not: `e = text.find(end)` (line 69) matches `<!-- uncoded:end -->` anywhere, including prose. All four failures trace to this locator plus the line-compare.

1. Unanchored end (line 69): a bare end marker in prose is taken as the boundary. `find` returns the earliest occurrence, so a replace spans from the real start to the prose marker, deleting the canonical body between. The exact mirror of the start-side fix.
2. CRLF (the compare, ~line 80): `existing_opening = text[s:line_end]` with `line_end = text.find("\n", s)` keeps the trailing `\r`, so it never equals `start` and always replaces; `after = …lstrip("\n")` does not strip `\r`, leaving mixed endings. Convergence takes an extra rewrite.
3. Duplicate sections: `re.search` returns only the first match (line 67); a second start-prefixed block is never located, so it is orphaned forever.
4. End-before-start: if the end marker precedes the start, `s < e` is false (line 70), so `section_found` is false and the code appends. Each sync adds a fresh section without removing the misplaced one; it never settles.

Root cause: one locator that is not symmetric with its start counterpart, plus a line-compare that is not newline-normalised. The fix is structural — make end-marker location line-anchored like the start, normalise line endings, and define a policy for duplicate/misordered markers.

Stamp derivation (GH173's test note): `MARKER_START` is `…sha256={hash(_CODE_SECTION_BODY)[:8]}…` (lines 35-38), computed at module load. Nothing unit-tests that the stamp derives from the body — only CI's `uncoded check` on committed files would catch a broken derivation.

## GH169 + source-read candidate — the reader siblings

`iter_source_files` (`extract.py:55`) wraps only `ast.parse` in try/except (catches `SyntaxError`, warns to stderr, continues). Its own `source = py_file.read_text()` is unguarded — a non-UTF-8 or unreadable `.py` file raises and aborts sync. `iter_doc_files` (`docs_map.py:41`) has no guard at all: both branches call `read_text(encoding="utf-8")` bare. So the "sibling already handles this" premise in GH169 is only half-true — the source sibling has the same latent read gap. The coherent fix gives both readers the same skip-and-warn on `UnicodeDecodeError`/`OSError`.

## GH167 + GH163 — `dispatch_rule.md` injected prose

This file is injected verbatim into every consumer repo. Two defects in the same prose: the worked examples name `resolve_body` and `function_name` (uncoded internals absent from consumer repos), and the paragraph repeats "is the rule firing" three times. Editing the file changes `_CODE_SECTION_BODY`, which changes the sha256 stamp, which changes `MARKER_START` — so this repo's own `CLAUDE.md`/`AGENTS.md` injected sections must be regenerated by sync (the dogfooding loop). A coherence point for development.

## GH175 — `_sync` build-or-remove shapes (`cli.py`)

Two shapes for one rule. Namespace map and stubs use `if config.source_roots: … else: remove…` (an explicit block). The skill uses `sync_skill(build=bool(config.source_roots))` (the branch pushed inside). Both correct; the divergence is what a future author trips over. Per the code-shape ladder, unifying the structure beats a prose comment — Design's call.

## GH174 — coverage gaps

`sync_skill`'s `build=False` arm (`skill.py`) is exercised only via CLI integration tests (`test_cli.py::TestSyncDocRoots`); `test_skill.py` covers only `build=True`. The fingerprint suite (`test_instruction_files.py::TestSyncInstructionFileFingerprint`) drives only the code section; the docs section runs the same `_apply_section` but no test drives it under reflow / different-stamp / plain-marker.

## GH168 + GH176 — docs

The pre-commit hook runs `uv run uncoded sync` (`.pre-commit-config.yaml`), regenerating `namespace.yaml`, stubs, `docs.yaml`, and the instruction-file sections — but no doc names this dogfooding (GH168). The hand-written Commands section (`CLAUDE.md` lines 39-64) tells readers to use `uvx uncoded …`, which runs the published package — wrong for a contributor working on local changes. README dev-setup (line 233) wires the hook but never states the `uvx` vs `uv run` distinction (GH176).

alimanfoo:
# Session Scope

Coherent Scope — the only option that applies. Nothing optional is left to strip for a Minimal (the two promoted candidates are requirements, the dogfooding regeneration is forced by CI, and the issue-named tests are required). The one wider candidate, an enforcing check for the build-or-remove rule, was deliberately rejected in #172, so there is no Maximal.

Each item states the property the work must achieve; the fix shape is Design's call.

- S1 (GH173). `_apply_section` is convergent and non-destructive: for arbitrary instruction-file contents, sync twice equals sync once, and no text outside the uncoded markers is ever deleted. All four reported failure modes are resolved. [Code Analysis: GH173]
- S2 (GH173). Convergence is guarded by tests: a property-based or equivalently thorough test asserts sync-twice-equals-once over arbitrary instruction-file contents; a unit test asserts that `MARKER_START`'s stamp derives from the section body. [Code Analysis: stamp derivation]
- S3 (GH169 + promoted candidate). Both readers degrade gracefully: `iter_doc_files` and `iter_source_files` each skip an unreadable or non-UTF-8 file with a one-line stderr warning and continue; one bad file never aborts the whole sync. [Code Analysis: reader siblings]
- S4 (GH167 + GH163). Injected dispatch-rule prose is consumer-safe and readable: `dispatch_rule.md` names no uncoded-only symbol and resolves the repeated "is the rule firing" phrasing, with meaning preserved; this repo's own `CLAUDE.md`/`AGENTS.md` injected sections are regenerated to match. [Code Analysis: dogfooding loop]
- S5 (GH175). One model for build-or-remove of code artefacts in `_sync` — the per-root-type pattern reads as one consistent shape. No enforcing check. [Code Analysis: GH175]
- S6 (GH174). Direct unit tests for the two under-covered paths: `sync_skill` with `build=False` (removes existing files; returns False when already absent; check mode reports without removing), and docs-section fingerprint tests paralleling the code-section ones.
- S7 (GH168). Dogfooding documented: a repo doc states that uncoded runs on itself via the pre-commit hook and what `uncoded sync` regenerates.
- S8 (GH176). Command guidance distinguishes consuming (`uvx uncoded`) from developing (`uv run uncoded`) in the Commands section of `CLAUDE.md`/`AGENTS.md` and in the README dev-setup section.

alimanfoo:
# Design

The design the session is building. Each item names the target shape; the fix shapes were weighed against the alternatives listed at the end.

## S1 — `_apply_section` (the GH173 fix)

Replace the two independent locators (`re.search` on the start marker, `text.find` on the end marker) with one line-oriented locator that anchors both markers to line starts. The locator finds the first start-prefix line, then the first end-marker line after it. Order-sensitivity then falls out for free: an end marker before any start cannot be read as a section (fixes end-before-start), and a stray end marker in prose before the real start is never mistaken for the boundary (fixes the prose-deletion case).

- Preserve the stamp tolerance: if the located opening line's stamp equals the canonical `MARKER_START`, leave the body untouched so a formatter's reflow survives; otherwise replace the located section with the canonical one.
- CRLF: compare marker lines with a trailing `\r` ignored, and write the canonical section with `\n`, leaving the surrounding user text — and its line endings — byte-for-byte. This converges in one pass and never rewrites content outside the markers.
- Duplicate or misordered markers: manage exactly the first well-formed section; collapse any additional well-formed uncoded section (a start-prefix line through its matching end line, distinctively ours) to keep one; leave a lone orphan marker untouched, so prose is never deleted.
- The removal path (`body is None`) uses the same locator.

## S2 — convergence tests

Add `hypothesis` as a dev dependency. A property test over arbitrary text asserts two laws: `apply(apply(t)) == apply(t)`, and every input line that is not one of the markers and not inside the managed section survives. Parametrised regression cases name the four known failure modes (CRLF, duplicate, misorder, prose end-marker). A unit test asserts that `MARKER_START` ends with `sha256(_CODE_SECTION_BODY)[:8]`, so a broken stamp derivation fails fast rather than only at CI's `uncoded check`.

A property-based test is chosen over hand-rolled cases alone because GH173 itself arose from missed enumeration, and the new duplicate/orphan policy has corners that are hard to pre-enumerate. The property test pins the whole class; the named cases ride alongside as regression documentation.

## S3 — graceful reads, one home

A shared helper reads a file's text and, on `UnicodeDecodeError` or `OSError`, prints a one-line `warning: skipping ...` to stderr and returns `None`. Both `iter_doc_files` and `iter_source_files` call it; `iter_source_files` keeps its own distinct `SyntaxError` message for the parse step. The guard lives in one place rather than two, because GH169 is itself a drift bug — the two readers diverged because nothing kept them in step, and a single home stops that recurring. `iter_source_files`' docstring is updated so its stated contract covers read-skips as well as the syntax-error skip.

## S5 — one build-or-remove model

Extract `_sync_code_artefacts(build, ...)` holding the namespace-map and stubs `if build: ... else: remove ...` block. `_sync` then reads as two parallel calls — `_sync_code_artefacts(build=...)` and `sync_skill(build=...)` — one shape, without changing the lower-level `build_stubs`/`sync_file` signatures that have other callers.

## S4 — injected prose

Edit `dispatch_rule.md`: replace the uncoded-only example symbols (`resolve_body`, `function_name`) with generic stand-ins, and rewrite the three "is the rule firing" sentences so the phrase is not repeated, with the meaning preserved. Then `uv run uncoded sync` regenerates this repo's injected `CLAUDE.md`/`AGENTS.md` sections — the stamp changes, so regeneration is forced. `CLAUDE.md` is a symlink to `AGENTS.md`, so this is one file.

## S6 — tests

Direct unit tests for `sync_skill` with `build=False` (removes existing files; returns False when already absent; check mode reports without removing), and a docs-section fingerprint test class paralleling the existing code-section one.

## S7 / S8 — docs

A dogfooding note in the README dev-setup section names the pre-commit hook and what `uncoded sync` regenerates. The `uvx uncoded` (consuming) versus `uv run uncoded` (developing against local source) distinction is added to the hand-written Commands section of `CLAUDE.md`/`AGENTS.md` and the README dev-setup. Both are developer guidance, legitimately prose — there is no code shape that carries "which command a contributor should type."

## Alternatives considered

- S1 line-anchored regex `^start$.*?^end$` (`MULTILINE|DOTALL`), de-duping via `re.sub`: concise, but correctness leans on regex semantics at file boundaries and trailing newlines, which both reviewers flagged as easy to misread. Set aside for the explicit scan.
- S2 hand-rolled cases only, no `hypothesis`: no new dependency and faster tests, but bounded by what the author enumerates — the exact limitation that produced GH173. Set aside.
- S5 thread `build=` into every sync function: also correct, but spreads the change across more files and signatures with other callers. Set aside.
- S3 duplicate the read-guard in both modules: avoids any shared dependency, but reinvites the very drift GH169 is about. Set aside for the single home.

alimanfoo:
# Plan

Seven tasks, one commit each, delivering the accepted Design.

- T1 (GH173). Rewrite `_apply_section` so it converges (sync twice equals sync once) and never deletes content outside the markers. One line-oriented locator anchors both markers to line starts: the first start-prefix line, then the first end-marker line after it. Preserve the stamp tolerance; compare marker lines ignoring a trailing `\r` and write the section with `\n` while leaving surrounding text byte-for-byte; collapse any extra well-formed section to one and leave a lone orphan marker untouched; the removal path uses the same locator. Update the function's docstring to match. Add `hypothesis` as a dev dependency. Tests: a property test asserting `apply(apply(t)) == apply(t)` and that every non-marker line outside the managed section survives; parametrised regression cases for CRLF, duplicate, misorder, and prose-end-marker; a unit test that `MARKER_START` ends with `sha256(_CODE_SECTION_BODY)[:8]`.
- T2 (GH169 + promoted candidate). Make both sync file-readers skip an unreadable or non-UTF-8 file with a one-line stderr warning and continue. Add a shared helper that catches `UnicodeDecodeError`/`OSError`, warns once, returns `None`; `iter_doc_files` and `iter_source_files` both call it. `iter_source_files` keeps its own `SyntaxError` message and gets its docstring updated to cover read-skips. Tests for each reader's skip-and-continue.
- T3 (GH175). Refactor, behaviour-preserving: extract `_sync_code_artefacts(build, ...)` holding the namespace-map and stubs build/remove block, so `_sync` reads as two parallel `build=` calls alongside `sync_skill(build=...)`. No change to the lower-level signatures. Existing `_sync` tests stay green.
- T4 (GH167 + GH163). Edit `dispatch_rule.md`: replace the uncoded-only example symbols with generic stand-ins, and rewrite the repeated "is the rule firing" sentences so the phrase isn't repeated, meaning preserved. Then regenerate the injected instruction-file sections with `uv run uncoded sync`.
- T5 (GH174). Direct unit tests: `sync_skill` with `build=False` (removes existing files; returns False when already absent; check mode reports without removing); a docs-section fingerprint test class paralleling the code-section one.
- T6 (GH176). Add the `uvx uncoded` (consuming) versus `uv run uncoded` (developing against local source) distinction to the hand-written Commands section and the README dev-setup section.
- T7 (GH168). Add a dogfooding note to the README dev-setup section naming the pre-commit hook and what `uncoded sync` regenerates.

Order: T1, T2, T3, T4, T5, T6, T7. No hard inter-task dependencies.

alimanfoo:
## Review — completeness and coherence against the agreed scope

**Recommendation: ready to merge. All agreed scope delivered; no coherence gaps.**

### Completeness against agreed scope

S1 (GH173) — `_apply_section` rewritten as a line-scan state machine anchoring both markers; CRLF handled; duplicate and misorder policies implemented; docstring updated. Delivered.

S2 (GH173) — Two Hypothesis property tests (idempotency; prose outside the section survives) plus five named regression cases and the stamp-derives-from-body unit test. Hypothesis added as a dev dependency. Delivered.

S3 (GH169 + candidate) — `read_helpers.py` adds `_read_file_text`; both `iter_source_files` and `iter_doc_files` call it and skip-and-continue on None; warning format is consistent (relative path in both iterators); both docstrings updated. Delivered.

S4 (GH167 + GH163) — `dispatch_rule.md` replaces `resolve_body` with `parse_config` and `function_name` with `validate_input`; three "is the rule firing" phrases rewritten without repetition; `AGENTS.md` regenerated with the new stamp; `CLAUDE.md` follows via symlink. The line-132 `function_name` (name-path syntax placeholder, not a symbol reference) correctly left unchanged. Delivered.

S5 (GH175) — `_sync_code_artefacts` extracted from `_sync`; `_sync` now shows two parallel `build=` calls for code artefacts. Delivered.

S6 (GH174) — Three `build=False` tests for `sync_skill` in `test_skill.py`; five docs-section fingerprint tests in `TestSyncInstructionFileFingerprintDocs` paralleling the code-section class. Delivered.

S7 (GH168) — README dev-setup paragraph names the pre-commit hook, lists what `uncoded sync` regenerates, and explains the re-stage step. Delivered.

S8 (GH176) — `AGENTS.md` Commands section and README dev-setup both distinguish `uvx uncoded` (published release) from `uv run uncoded` (local working tree). Delivered.

### Whole-diff coherence

Nothing additional needed. The changes fit together cleanly across commits: the dogfooding loop in the dispatch-rule task ran `uv run uncoded sync` through the fixed `_apply_section`, producing the correct stamp transition (`5d50217d` to `d50048f1`). The new `_read_file_text` and `_sync_code_artefacts` are both correctly indexed in `namespace.yaml` and their stubs. The two README paragraphs are adjacent and flow naturally as a unit. All 394 tests pass at 100% branch coverage.

**Out of scope but noticed**: nothing.

alimanfoo:
## Review — fresh-eyes read

**Recommendation:** Looks good — clean, well-tested maintenance work; a few small notes, nothing blocking.

I built my cold read from the diff first, then compared against the PR description. The two matched closely: the one spot my read couldn't explain on its own (the motivation for the `_apply_section` rewrite) is exactly GH173, and it's well documented. The `cli.py` extraction is a faithful behavior-preserving refactor. `_apply_section` is idempotent across the cases I traced — CRLF, duplicates, orphan markers, end-before-start, empty input — and never deletes prose outside the markers.

### Non-blocking

1. **Silent UTF-8 narrowing in `extract.py`.** `iter_source_files` previously called `py_file.read_text()` with no encoding, which uses the platform locale; it now reads UTF-8 via `_read_file_text`. A `.py` file with a PEP 263 declaration like `# -*- coding: latin-1 -*-` that read fine before is now treated as non-UTF-8, skipped, and dropped from the index with only a warning. This is the right call for a UTF-8-first tool, but the change is framed everywhere as "skip unreadable files" — neither the docstring nor the commit notes that readable-file behavior also narrows. A one-line note in the docstring would close the gap.

### Nits

1. **CRLF separator on first injection.** In `_apply_section`, the append and orphan-insert paths use `rstrip("\n")`, not `rstrip("\r\n")`. When a section is first injected into a CRLF file with no prior section, the blank separator line is emitted as LF, mixing line endings. The marker-match path handles CRLF; this first-injection path still mixes. Low severity since the canonical section is LF regardless.
2. **Inconsistent sibling warnings.** In `iter_source_files`, the read-skip path prints `warning: skipping {path}: {raw exception}` while the syntax-error path prints a structured `warning: skipping {path}: SyntaxError at line N: msg`. Same iterator, two formats.
3. **String-literal state machine.** `_apply_section` drives its scan on bare strings `"before"`/`"in_first"`/`"after"`/`"in_extra"`. A typo in any branch wouldn't be caught by tooling. Contained and well-tested, but named constants or an enum would make a future edit safer.

### Readability

1. **`_apply_section` docstring is accurate but dense.** Five labeled paragraphs plus a four-case bullet list for one function. Leading with the one-sentence guarantee — idempotent, and never deletes text outside the markers — before the per-case detail would let a reader confirm the essential property fast.

alimanfoo:
## Response to the reviews

Thanks to both reviewers. The completeness-and-coherence review found all agreed scope delivered with no gaps and nothing further needed. The fresh-eyes review approved with a few small notes, handled as follows.

**Accepted**

- *UTF-8 narrowing under-documented.* `iter_source_files` now reads source as UTF-8, so a file with a non-UTF-8 encoding declaration is skipped — behaviour the docstring framed only as "fails to read". Documented in commit 4bd7f40, with the latin-1 example.
- *`_apply_section` docstring dense.* The docstring now leads with the core guarantee — idempotent, and never deletes text outside the markers — before the per-case detail. Commit 4bd7f40.

**Rejected**

- *CRLF separator emitted as LF on first injection.* Convergence holds: the idempotency property test passes over arbitrary text, including CRLF. The separator is part of the canonical section, which is written with `\n` by design; making a CRLF file fully consistent would mean rewriting the section itself to CRLF, which the design deliberately does not do. Left as is.
- *Read-skip and syntax-error warnings format differently in the same iterator.* The two report different error classes: the syntax error carries a structured line and message, while the read error's raw message is adequate. The difference is intentional, not an oversight.
- *State machine driven by string literals.* The states are local to one function with 100% branch coverage; a typo'd state drops a branch and fails the coverage gate. An enum would add machinery without buying safety here.

src/uncoded/instruction_files.py
```diff
@@ -53,43 +52,102 @@ def _apply_section(
...
     When body is a string:
     - absent → append the canonical section;
-    - found and opening-marker line matches start → return text unchanged
-      (a formatter's reflow of the body is tolerated);
-    - found and opening-marker line differs → replace the whole section.
-    When body is None: remove the section if present.
+    - found and opening-marker line matches start → managed section preserved
+      byte-for-byte; any extra copies collapsed;
+    - found and opening-marker line differs → replace with the canonical section.
+    When body is None: remove the section and any extra copies if present.
```
alimanfoo, line 83: This complexity is a smell. It makes me want to rethink the whole policy of injecting instructions into AGENTS.md/CLAUDE.md.

src/uncoded/read_helpers.py
```diff
@@ -0,0 +1,23 @@
+"""File-reading helpers shared by sync iterators."""
+
+import sys
+from pathlib import Path
+
+
+def _read_file_text(path: Path, *, display: Path | str | None = None) -> str | None:
```
alimanfoo: Perhaps this should be called something like "_read_file_text_as_utf8" so less comments are needed around call sites, the name is more self-explaining.

Issue #183, opened 2026-06-19
Explore and Plan subagents never receive uncoded's navigation guidance


alimanfoo:
Noticed during #177, while reconsidering #123. uncoded's navigation guidance reaches agents through the injected section in `AGENTS.md`/`CLAUDE.md`. Per the Claude Code [sub-agents docs](https://code.claude.com/docs/en/sub-agents), the built-in Explore and Plan subagents skip `CLAUDE.md` (and don't preload skills), so they never see the dispatch rule or the instruction to load `namespace.yaml`.

Explore is the subagent Claude delegates codebase searching to — exactly the work the "use the index, not grep" rule is meant to steer. So the one agent most likely to grep is the one the guidance can't reach.

This is independent of how the guidance is packaged: an injected section, a separate file, or a skill are all reached the same way, and Explore/Plan skip all of them. The docs note the standard mitigation is to restate a must-have rule in the delegation prompt when handing work to a built-in subagent.

Suggested direction: investigate what, if anything, can reach Explore — restating the rule at delegation time, or another mechanism — and document the limitation either way.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #178, opened 2026-06-19
File-read encoding handling is inconsistent and partly unguarded


alimanfoo:
Noticed during #177. uncoded's file reads handle encoding inconsistently, and one site #80 named is still unfixed.

- `sync_instruction_file` reads the instruction file with a bare `target.read_text()` (`src/uncoded/instruction_files.py:178`) — no explicit encoding and no error handling. A corrupt or non-UTF-8 `AGENTS.md`/`CLAUDE.md` gives an unhandled traceback instead of a clean `Error:` and exit 1. #80 listed this exact site to fix; the fix never reached it.
- #177 made the source and doc readers read UTF-8 explicitly and skip unreadable files. Correct for a UTF-8-first tool, but it now silently drops a valid Python file that declares a non-UTF-8 encoding (e.g. `# -*- coding: latin-1 -*-`).

Suggested direction: give `sync_instruction_file`'s read the guard pattern the rest of `_sync` uses (clean error, exit 1); and consider `tokenize.detect_encoding` (stdlib) so source is read under its declared encoding rather than forced to UTF-8. Related: #80, #169.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #184, opened 2026-06-19
Ship navigation guidance as skills, drop AGENTS.md/CLAUDE.md injection

### LINKED ISSUE #123: Migrate from AGENTS.md to a skill
Possible follow-on from #111 would be to migrate the agent instructions from AGENTS.md to a skill. This would make the uncoded behaviour opt-in, i.e., users could choose to run their agents without it if they preferred.

### LINKED ISSUE #181: _apply_section repairs malformed instruction-file markers silently
Noticed during #177. `_apply_section` (`src/uncoded/instruction_files.py`) now silently collapses duplicate uncoded sections and inserts the canonical section before a lone orphan start marker. Other managed-block tools (Ansible managed blocks, ssh `authorized_keys` managers) surface malformed markers rather than repairing them invisibly, so the user learns their file was malformed.

Suggested direction: print a one-line stderr warning when a repair happens — for instance when duplicate sections are collapsed. Scope it to the repair cases; a left-untouched orphan should not warn on every sync, or the warning becomes persistent noise. Lower priority.

### LINKED ISSUE #182: Injected docs guidance doesn't tell agents to load docs.yaml
Noticed during #177. The injected code guidance tells an agent to read `.uncoded/namespace.yaml` in full at session start ("Orient … before any other tool call"), so the symbol index is in context from the outset. The injected docs guidance (`src/uncoded/docs_rule.md`) only describes `.uncoded/docs.yaml` and says to navigate it with `Read`/`grep` — it never tells the agent to load it up front.

The feature's stated intent (README: "Agents load this to orient to the documentation, then navigate to a heading") is for the docs index to be loaded into context the way the namespace map is. Without an orient instruction the docs index's main purpose goes unrealized.

Suggested direction: add an orient step to `docs_rule.md` parallel to the code side — tell the agent to read `docs.yaml` at session start. Related: #165.

alimanfoo:
This addresses #123. uncoded today injects its navigation guidance into a delimited section inside the consumer's hand-edited `AGENTS.md`/`CLAUDE.md`, and maintains that section on every `sync`. This change removes that injection and ships the guidance as generated skill content instead, alongside a one-line pointer a maintainer adds by hand if they want the guidance always loaded.

This is a breaking change: it removes the `instruction-files` config key and stops uncoded writing into `AGENTS.md`/`CLAUDE.md`, so host repos configure differently. It ships as v2.0.0 (the version comes from the git tag).

## Improvement goals

1. **uncoded stops owning a section inside hand-edited `AGENTS.md`/`CLAUDE.md`.** The injection mechanism is removed: `instruction_files.py`, the `MARKER_*` constants, `_apply_section`'s line-scan state machine (CRLF, duplicate sections, orphan markers, stamp/reflow tolerance), and the `instruction-files` config key. `sync` writes no instruction-file sections.

2. **Navigation guidance ships as generated skill content instead.** Skill generation becomes a small data-driven registry. Three skills ship, each named with an `uncoded-` prefix for provenance and collision-safety: `uncoded-code-navigation` (when `source-roots` is set), `uncoded-doc-navigation` (when `doc-roots` is set), and the renamed `uncoded-coherence-review`. Each is generated the same generate-or-noop way and written for both `.claude/skills/` and `.agents/skills/`. The old `coherence-review` skill is migrated automatically.

3. **Delivery is on-demand by default, always-on by opt-in.** The guidance is available on demand through the skill mechanism. A maintainer who wants it always loaded adds a one-line instruction to `CLAUDE.md`/`AGENTS.md` by hand, once. uncoded no longer writes that line.

4. **Close #181.** #181 asks `_apply_section` to stop silently repairing malformed markers. Removing `_apply_section` entirely resolves it.

5. **Fix #182.** The docs-navigation guidance now tells an agent to load `docs.yaml` at session start, parallel to the code side.

## Preserved behaviour

- Guidance still reaches agents, including custom and general-purpose subagents — on demand via the Skill tool, and via the manual always-on pointer for `CLAUDE.md`-reading subagents.
- Per-root-type independence: code guidance is produced when `source-roots` is set, docs guidance when `doc-roots` is set, each independent of the other.
- Cross-agent support: guidance is written for both the Claude (`.claude/skills/`) and Codex (`.agents/skills/`) populations.
- `uncoded check` still detects drift of everything uncoded owns — now the skill files — for CI.
- The index artefacts (`namespace.yaml`, stubs, `docs.yaml`) and the `body`/`refs` tools are untouched.

## Migration

The README carries an "Upgrading from v1" guide: remove the old `<!-- uncoded:start … -->` / `<!-- uncoded:docs:start … -->` marker blocks by hand, update any manual `coherence-review` pointer to `uncoded-coherence-review`, and drop the now-inert `instruction-files` config key. The README also shows a first-time always-on example. uncoded's own `AGENTS.md`/`CLAUDE.md` and README are migrated to the new model as part of dogfooding.

## Out of scope

- The Explore/Plan subagent blind spot — those built-ins skip `CLAUDE.md` and skills alike. It limits every approach equally and is tracked separately in #183.

Closes #123
Closes #181
Closes #182


alimanfoo:
# Code Analysis

A structural read of how uncoded's agent-facing guidance is produced today, and the cascade the removal touches.

## The core pattern: injection is the lone exception to the generate-or-noop model

uncoded produces every artefact one way except one. The namespace map, the stubs, `docs.yaml`, and the coherence-review skill are all generated wholesale and written through `sync_file`/`remove_file` (`src/uncoded/sync.py`) — write only when content differs, `check` mode reports without touching disk, removal is a clean `unlink`. uncoded owns the whole file; nothing else edits it.

The instruction-file sections break that model. uncoded owns only a delimited region inside `AGENTS.md`/`CLAUDE.md`, files humans and formatters also edit. Maintaining that region is `_apply_section` (`instruction_files.py:50-150`) — a 100-line line-scan state machine with four states, handling CRLF (`:95`, `:143`), duplicate-section collapse (`:129-138`), lone-orphan-marker insertion (`:116-124`), and a sha256 stamp/reflow tolerance scheme (`:25-47`, `:143-148`) so a formatter's rewrap survives. This complexity is the cost of owning a region in a hand-edited file — and removing that ownership is the point of the session.

## Recurrence confirms it

The record shows three issues landing on this one surface across sessions: #171 (formatter reflow fight, added the stamp tolerance), #173 (four convergence/corruption bugs in `_apply_section`), and #181 (silent marker repair, still open). That run of fixes on a single surface is not unrelated defects but one design choice — in-place section surgery on a shared file — generating defect after defect. Removing the mechanism retires the whole class.

## Removal cascade

The injection mechanism has a tight, fully-mapped footprint. Everything below exists only to serve it:

- **`src/uncoded/instruction_files.py`** — the whole module: `_apply_section`, `sync_instruction_file`, `SECTION_CODE`/`SECTION_DOCS`, `_CODE_SECTION_BODY`/`_DOCS_SECTION_BODY`, and all `MARKER_*` constants (`:18-150`).
- **`src/uncoded/cli.py:189-220`** — the instruction-section block in `_sync`: building the two sections, and the per-path loop with its symlink dedup (`:193-220`). That dedup exists only because two instruction files may be the same inode; it goes with the loop.
- **`src/uncoded/config.py`** — `Config.instruction_files` (`:21`), the `instruction-files` key read (`:122-126`), and the `DEFAULT_INSTRUCTION_FILES` import (`:7`). The field and key have no other consumer.
- **Tests** — `tests/test_instruction_files.py` (entire file); the instruction-file cases in `test_cli.py` (`test_writes_namespace_map_stubs_and_instruction_file`, `test_error_when_instruction_file_outside_project_root`, `test_check_returns_one_when_instruction_file_drifts`, the two symlink-dedup cases); and the config cases (`test_instruction_files_default_when_key_absent`, `test_instruction_files_configured`, `test_instruction_files_empty_list_respected`).

## Where the guidance lands instead — the skill model

The target model already exists in `src/uncoded/skill.py`. `sync_skill` writes `_SKILL_CONTENT` (loaded via `importlib.resources.files("uncoded")`) to `SKILL_OUTPUTS` — `.claude/skills/<name>/SKILL.md` (Claude) and `.agents/skills/<name>/SKILL.md` (Codex) — and cleans `LEGACY_SKILL_OUTPUTS`. Skill files carry YAML frontmatter (`name`, `description`); see `coherence_review.md:1-4`. The two guidance bodies that move are `dispatch_rule.md` (code navigation) and `docs_rule.md` (docs navigation), today injected verbatim — each has a single home already, so there is no duplicated-fact problem.

Two structural notes for the design phase (named, not decided here):

- **Per-root-type coupling.** `sync_skill` today takes one `build` flag, wired to `source_roots` (`cli.py:137,148`, with the comment "the skill depends on namespace.yaml and stubs, so it follows the same rule"). The new code-navigation skill keys off `source_roots`; the docs-navigation skill must key off `doc_roots` — so the single `build` flag no longer covers both.
- **#182 is a content gap in `docs_rule.md`.** It describes `docs.yaml` but never tells the agent to load it at session start, unlike `dispatch_rule.md:33-42` for the namespace map. The fix lives in the migrated docs-skill content.

## Architecture and conventions touched

- **Generate-or-noop is the convention; injection is the exception.** This change deletes the exception and brings the guidance under the convention. Nothing enforces the convention — #175 considered an enforcing check and deliberately rejected it; it stays convention-only.
- **Per-root-type build/remove** in `_sync`: code artefacts when `source_roots` set, doc artefacts when `doc_roots` set. The new skills must honour this.
- **Packaging.** The guidance `.md` files ship as package data under `src/uncoded/` and are read via `importlib.resources`; hatchling includes them by default (no explicit manifest). New skill content files follow the same path.
- **Dogfooding fallout.** uncoded's own `AGENTS.md`/`CLAUDE.md` carry the injected sections today, and README documents both the injection (`README.md:40,88,174`) and a symlink-drift caveat tied to it (`README.md:269-278`). All of that migrates: the sections become a manual pointer, and the symlink-drift caveat — which exists because uncoded writes into those files — largely evaporates.

alimanfoo:
# Session Scope

The Coherent Scope, adopted. This is a maintenance session: coherence means fixing every instance of the inconsistency — the whole injection exception, not just `_apply_section`.

1. **Remove the injection mechanism** — uncoded writes no section into `AGENTS.md`/`CLAUDE.md`; `_apply_section`, the markers, and the stamp/reflow scheme no longer exist.
2. **Remove the dead config surface** — the `instruction-files` key, `Config.instruction_files`, and `DEFAULT_INSTRUCTION_FILES`.
3. **Deliver the code- and docs-navigation guidance as generated skill content**, under the generate-or-noop model, written for both the Claude and Codex skill locations.
4. **Honour per-root-type independence** — code guidance produced when `source-roots` is set, docs guidance when `doc-roots` is set.
5. **Docs guidance tells agents to load `docs.yaml` at session start** (fixes #182).
6. **`uncoded check` detects drift of the new skill content.**
7. **Tests** — remove the now-dead injection tests; cover the new skill-delivery and per-root-type paths to the repo's 100% branch-coverage gate.
8. **uncoded dogfoods the new model** — its own `CLAUDE.md`/`AGENTS.md` drop the injected sections in favour of the manual pointer, and README is updated to document skills plus the manual pointer and to drop the injection docs, including the symlink-drift caveat that injection caused.

Outcome: closes #123, #181 (resolved by removal), #182 (content fix).

Not added: nothing enforces the generate-or-noop convention. #175 deliberately rejected such a check and that decision stands, so this session adds none.

No Maximal scope: the adjacent open issues #183 (Explore/Plan reach) and #178 (source-file encoding) are independent of how the guidance is packaged, not forward-leads of this removal. No narrower Minimal cut: removing every instance is the maintenance definition here.

alimanfoo:
# Design

Skills as data — a small registry replaces the section-injection mechanism.

## `skill.py` — a registry

- A frozen `Skill` dataclass: `name`, `description`, `body_file` (a package resource holding body content only), `gate` (`Literal["code", "docs"]`), and `legacy_names: tuple[str, ...] = ()`.
- One constant for the output locations: `SKILL_ROOTS = [.claude/skills, .agents/skills]`. A skill's output path is `root/<name>/SKILL.md` for each root, so the Claude-and-Codex two-location fact lives in one place (today it is written out twice).
- A `SKILLS` list: `uncoded-coherence-review` (gate `code`; legacy names `coherence-review`, `uncoded-review`), `uncoded-code-navigation` (gate `code`), `uncoded-doc-navigation` (gate `docs`).
- A driver `sync_skills(*, source: bool, docs: bool, project_root, check) -> int`: for each skill, build it when its gate is satisfied, otherwise remove it; always remove the `legacy_names` paths. The written content is the frontmatter rendered from the row (`name`, `description`) followed by the body file. Routes through `sync_file`/`remove_file`, so check-mode and content-aware writes come for free.

Generating the frontmatter from the row single-sources the skill name: it would otherwise live both in `Skill.name` (which drives the output directory) and in the file's `name:` frontmatter (which drives invocation), with nothing enforcing the match.

## CLI and config

- `cli._sync`: replace the instruction-file block and the single skill call with one `sync_skills(source=bool(source_roots), docs=bool(doc_roots), …)`. Drop the `instruction_files` import. Update the `sync` subcommand `--help` text, which today names "instruction-file sections".
- `config.py`: remove the `instruction_files` field, the `instruction-files` key read, and the `DEFAULT_INSTRUCTION_FILES` import. A leftover `instruction-files` key in a consumer's config is then inert — read by nothing, no error. The migration guide notes this.
- `sync.py`: the module docstring lists "instruction-file sections" as an artefact type; update it to "skill files".

## Deletions

- `src/uncoded/instruction_files.py` — the whole module.
- `tests/test_instruction_files.py` — the whole file.

## Skill content

`dispatch_rule.md` becomes `code_navigation.md` and `docs_rule.md` becomes `doc_navigation.md`, each holding body content only. `coherence_review.md` is stripped to body only. The docs body gains an orient step telling the agent to load `.uncoded/docs.yaml` at session start, parallel to the code side (closes #182). Each is lightly revised so it reads as a skill rather than an injected section.

## README and dogfooding

uncoded ships at v2.0.0 — removing the `instruction-files` key and the writes into `AGENTS.md`/`CLAUDE.md` changes how host repos configure, so it is a breaking change. The version comes from the git tag (`hatch-vcs`); the general release note goes on the GitHub release, not the repo.

README changes: replace the injection description with the skills model; add a migration guide (existing users remove the old `<!-- uncoded:start -->…<!-- uncoded:end -->` marker blocks by hand) and a first-time example showing the one-line instruction to add to `AGENTS.md`/`CLAUDE.md` for always-on loading of the two navigation skills; drop the symlink-drift caveat, which only existed because uncoded wrote into those files. uncoded's own `CLAUDE.md`/`AGENTS.md` drop their injected sections, and `check` stays green.

## Alternatives considered

- **Generalized helper** — `sync_skill(name, content, legacy, build)` called three times from the CLI. Delivers the full scope, but the set of skills lives across three call sites rather than one home, and adding a skill means touching the CLI.
- **Thin per-skill functions** — one near-identical build/remove function per skill. Simplest diff, but roughly three times the duplication and no shared shape.
- **Combined single skill** — one navigation skill covering both code and docs. Set aside: it breaks per-root-type independence, so an agent on a docs-only repo would load code-navigation content it cannot use.
- **Transitional auto-removal / `check`-warns on stale markers** — clean up existing consumers' old sections automatically. Set aside: both keep marker-recognition logic alive, which is the complexity this change exists to remove. The v2 migration guide covers the transition instead.

alimanfoo:
# Plan

Five tasks, ordered so every commit leaves the tree green and the dogfooded index in sync: build the registry, add the new skills, rename, then remove the old mechanism, with docs last.

1. **Refactor skill generation into a data-driven registry** (behaviour-preserving). Add a `Skill` dataclass (name, description, body_file, gate, legacy_names), a `SKILL_ROOTS` constant for the two output locations, a `SKILLS` list (coherence-review only at this point, name unchanged), and a `sync_skills` driver that builds or removes each skill by its gate and writes generated frontmatter plus the body file. Strip `coherence_review.md` to body only; point `_sync` at `sync_skills`; remove the now-stale coupling comment in `cli.py`. The coherence-review skill installs with the same name, description, body, and two locations as before.

2. **Add the code- and docs-navigation skills.** Create `code_navigation.md` and `doc_navigation.md` as body-only skill content; the docs body gains an orient step telling the agent to load `.uncoded/docs.yaml` at session start, parallel to the code side (closes #182). Add two registry rows, `uncoded-code-navigation` (code gate) and `uncoded-doc-navigation` (docs gate).

3. **Rename coherence-review to uncoded-coherence-review.** Change the registry name and add the old names to `legacy_names` so sync removes the old directories. All three skills now share the `uncoded-` prefix.

4. **Remove the injection mechanism.** Delete `instruction_files.py` and its tests; remove the instruction-file block, the import, and the stale `--help` text from the CLI; remove the `instruction_files` field, the `instruction-files` key read, and the default-files import from config; update the `sync.py` module docstring; delete the now-unused `dispatch_rule.md` and `docs_rule.md`; drop the injection-specific test cases. No caller or test references the injection symbols.

5. **Update README and dogfood the new model.** README documents the skills model, a migration guide (remove the old marker blocks by hand, and update any manual `coherence-review` pointer to `uncoded-coherence-review`), and a first-time example of the one-line always-on instruction; the symlink-drift caveat is dropped. uncoded's own `CLAUDE.md`/`AGENTS.md` drop their injected sections and gain the pointer.

alimanfoo:
## Review — completeness and coherence against the accepted scope

**Approve — the PR delivers the agreed scope in full with no gaps remaining.**

Reviewed the complete diff (commits e9f14e6 through 6ae2822) against the accepted requirements and Session Scope.

### Completeness against scope

All eight scope items are met.

1. **Injection mechanism removed.** `instruction_files.py` deleted, the `_apply_section` state machine and all markers gone, the cli.py injection block deleted, `instruction_files` import removed from cli.py. No injection symbol survives anywhere in src/ or tests/.
2. **Dead config surface removed.** `Config.instruction_files`, the `instruction-files` key read, and the `DEFAULT_INSTRUCTION_FILES` import are all gone from config.py. A leftover `instruction-files` key in a consumer's config file is silently ignored.
3. **Skills deliver code- and docs-navigation guidance, generate-or-noop.** `skill.py` is now a registry: `Skill` frozen dataclass, `SKILL_ROOTS` as the single home for the two output locations, `SKILLS` list with three entries, `sync_skills` driver, `_render_content` (frontmatter from registry row + lstrip body), `_remove_skill_file` (remove + prune empty parent directory). All three skills written to `.claude/skills/` and `.agents/skills/`.
4. **Per-root-type independence.** `uncoded-code-navigation` has `gate="code"`; `uncoded-doc-navigation` has `gate="docs"`; `uncoded-coherence-review` has `gate="code"`.
5. **Docs guidance loads docs.yaml at session start (#182).** `doc_navigation.md` step 1: "Read .uncoded/docs.yaml … Read it once, in full, at session start." #182 closed.
6. **`check` detects skill drift.** `sync_skills` routes through `sync_file`; `_sync(check=True)` exits non-zero if any change is detected.
7. **Tests to 100% branch.** 352 tests, 100% branch coverage across all task commits.
8. **README + dogfooding.** README replaces the injection description with the skills model; migration guide covers all three user steps; symlink-drift caveat removed; dev-setup paragraph updated. AGENTS.md/CLAUDE.md drop the injected marker blocks and carry the always-on pointer. docs.yaml regenerated.

Additional scope: #181 resolved by deletion of `_apply_section`. All three skills share the `uncoded-` prefix. `coherence-review` migrated to `uncoded-coherence-review` via `legacy_names`. Migration guide present for v2.0.0.

No gaps. The `docs_rule.md` → `doc_navigation.md` transition is correct; `sync_skills` ordering is immaterial (skill content is static); the `#coherence-review` README anchor targets the section heading; `resolved_project_root` is still live; `_DOC_SKILL_PATHS` is used through the combined list.

### Out of scope but noticed

Nothing worth filing.

alimanfoo:
## Review — fresh-eyes read

**Recommendation:** looks good — no blockers; the mechanism is correct and well-tested, with a couple of documentation gaps and small clarity items.

The rewrite from marker-injection to a data-driven skill registry is clean. I traced the check/apply paths closely: `uncoded check` predicts the same change count as `uncoded sync` in every branch, the empty-directory pruning is correctly skipped in check mode without causing a count divergence, and the loop is idempotent. The `bool → int` return change on `sync_skills` is arithmetically safe and the exit-code gate still works. No stale references to the removed mechanism survive anywhere — source, tests, the `.uncoded/` index, pyproject, or pre-commit config.

### Non-blocking

1. The "Upgrading from v1" guide tells consumers to delete the injected `<!-- uncoded:start -->` blocks, but never tells them to add the replacement skill pointer. The injected blocks *were* how navigation guidance reached agents in v1. A consumer who follows the three steps faithfully ends up with no always-on navigation guidance and no signal anything changed. The skill-pointer line exists in README's "what it generates" section but is framed as optional and isn't linked from the upgrade guide. Add a step pointing there, and name the on-demand-by-default shift.
2. The README dropped the "Note for Windows contributors", but `CLAUDE.md` is still a symlink to `AGENTS.md`. The gotcha hasn't gone away — on Windows without `core.symlinks`, `CLAUDE.md` is checked out as a plain file containing the literal text `AGENTS.md`. The old note's stated symptom ("the navigation section drifts") is obsolete, but the symlink fragility itself remains. Consider keeping a trimmed note.
3. Low priority: `_render_content` configures `yaml.dump` independently, while `yaml_tree.py` owns the project's other YAML rendering via a shared `_CleanDumper`. The needs genuinely differ, so full unification would be a poor fit — but there are now two places that pin PyYAML choices. A one-line comment saying why frontmatter doesn't reuse `render_yaml_tree` would close the question for the next reader.

### Nits

1. The path shape `root / skill.name / "SKILL.md"` is built three times in `skill.py` and again as comprehensions in `tests/test_cli.py`. A small `_skill_path(root, name)` helper would name the shape and make the active-vs-legacy distinction self-evident.
2. `sync_file` and `remove_file` are annotated `-> bool`, but `sync_skills` now sums their returns into an int count (relying on bool being an int subtype). It works, but the count-addable contract is invisible.
3. `build = source if skill.gate == "code" else docs` silently maps any future third gate value to `docs`. Fine for the two current gates; if a third is ever added, make it an explicit branch.

### Readability

My cold read of the diff matched the PR intent closely — the change is reviewable. The one reader-facing gap is the upgrade guide (Non-blocking 1): the shift from "guidance is always injected" to "guidance is an on-demand skill" is the consumer's biggest mental-model change, and the upgrade steps don't name it.

alimanfoo:
## Response to the reviews

The completeness-and-coherence review raised no findings — no response work needed there.

The fresh-eyes review raised non-blocking items and nits, handled as follows.

**Accepted and addressed**

- **Upgrade guide didn't name the model shift or how to keep always-on guidance** — fixed in d4270bc. The "Upgrading from v1" guide now states that guidance moves from always-on injection to on-demand skills, and adds a step to restore always-on via the pointer line.
- **Dropped Windows note** — fixed in d4270bc. A trimmed "Note for Windows contributors" is restored, covering the `CLAUDE.md` → `AGENTS.md` symlink checkout fragility, without the obsolete drift wording.
- **Path shape built inline three times** — fixed in 26c1de2. A `_skill_path(root, name)` helper now single-sources `root/<name>/SKILL.md`, making the active-vs-legacy distinction self-evident.

**Rejected, with reasons**

- **Comment explaining the `yaml.dump` divergence** — the frontmatter renderer and the tree renderer serve genuinely different needs (a flat two-key mapping vs a tree), and the divergence is self-evident from the code; a comment guarding a theoretical drift defends no behaviour.
- **`sync_file`/`remove_file` returning `int`** — summing their boolean returns into the change count is the established pattern across `_sync`, pre-dating this change; the functions correctly answer a yes/no question and changing their signatures is out of scope.
- **Raising `else` for a future third gate value** — `gate` is typed `Literal["code", "docs"]`, so a third value is a type error caught by the type checker; a runtime guard for a type-forbidden case would be speculative defensive code.

.agents/skills/uncoded-code-navigation/SKILL.md
```diff
@@ -0,0 +1,110 @@
+---
+name: uncoded-code-navigation
+description: 'Navigate Python source code in a codebase indexed by uncoded. Load .uncoded/namespace.yaml first, read .pyi stubs before source files, and apply the dispatch rule: symbol name to uncoded body/refs, pattern or phrase to grep.'
```
alimanfoo: The description should say when the skill should be applied, it's a trigger. Less what the skill mechanically does.

.agents/skills/uncoded-doc-navigation/SKILL.md
```diff
@@ -0,0 +1,22 @@
+---
+name: uncoded-doc-navigation
+description: Navigate a codebase's Markdown documentation indexed by uncoded. Load .uncoded/docs.yaml at session start to see every file and its heading hierarchy, then use Read or grep to reach a specific section.
```
alimanfoo: Mechanical details not relevant here.

.agents/skills/uncoded-code-navigation/SKILL.md
alimanfoo: Fixed in 38110bb. The description is now a trigger — "Use when finding, reading, or editing Python source in a codebase indexed by uncoded — locating a symbol, reading a definition, or checking references before a refactor, rename, or delete." The mechanical detail (load namespace.yaml, read stubs, dispatch rule) stays in the skill body where it belongs.

.agents/skills/uncoded-doc-navigation/SKILL.md
alimanfoo: Fixed in 38110bb. Now "Use when finding or reading a codebase's Markdown documentation indexed by uncoded — locating which file and section cover a topic, or orienting to what documentation exists." The load-docs.yaml-at-session-start mechanics are dropped from the description and remain in the skill body.

.agents/skills/uncoded-code-navigation/SKILL.md
```diff
@@ -0,0 +1,110 @@
+---
+name: uncoded-code-navigation
+description: Use when finding, reading, or editing Python source in a codebase indexed by uncoded — locating a symbol, reading a definition, or checking references before a refactor, rename, or delete.
```
alimanfoo:
```suggestion
description: Use before searching, reading, or editing Python source in a codebase indexed by uncoded — locating a symbol, reading a definition, or checking references before a refactor, rename, or delete.
```

.agents/skills/uncoded-code-navigation/SKILL.md
```diff
@@ -0,0 +1,110 @@
...
+---
+
+# Code Navigation
+
+This codebase uses [uncoded](https://github.com/alimanfoo/uncoded) to maintain
+a symbol index over its source code, with two associated tools:
```
alimanfoo:
```suggestion
a symbol index over its source code, with two associated CLI tools:
```

.agents/skills/uncoded-code-navigation/SKILL.md
```diff
@@ -0,0 +1,110 @@
...
+`uncoded body` for reading a symbol's body and `uncoded refs` for finding
+references.
+The point of this scaffolding is one rule.
```
alimanfoo:
```suggestion
```

.agents/skills/uncoded-doc-navigation/SKILL.md
```diff
@@ -0,0 +1,22 @@
+---
+name: uncoded-doc-navigation
+description: Use when finding or reading a codebase's Markdown documentation indexed by uncoded — locating which file and section cover a topic, or orienting to what documentation exists.
```
alimanfoo:
```suggestion
description: Use before searching or reading a codebase's Markdown documentation indexed by uncoded — locating which file and section cover a topic, or orienting to what documentation exists.
```

.agents/skills/uncoded-doc-navigation/SKILL.md
```diff
@@ -0,0 +1,22 @@
...
+
+**Step 1 — Orient. Read the docs map first.** Before answering the
+user, before any other tool call:
+
+```text
+Read .uncoded/docs.yaml
+```
+
+`.uncoded/docs.yaml` is an orientation outline: it lists every Markdown file
+and its heading hierarchy. Read it once, in full, at session start.
```
alimanfoo:
```suggestion
`.uncoded/docs.yaml` is an orientation outline: it lists every Markdown file
and its heading hierarchy. Read it in full, now.
```

alimanfoo (commented):
See line comments...

PR #186, opened 2026-06-23
Add docstring style checker

### LINKED ISSUE #166: Add docstring style checker
Add a docstring style checker to ensure consistent use of docstrings across Python code in the project. Choice of style and linter TBD. Adding this check would also require fixing docstrings which don't conform, of course.

alimanfoo:
Adds an automated docstring style check to the project's tooling and brings existing docstrings into conformance. Addresses #166.

### Improvement goals

1. Every public Python symbol (module, class, function, method) carries a docstring conforming to the pep257 convention, enforced by an automated check that fails on violation in both pre-commit and CI. The rule then holds without anyone having to remember it.
2. Existing docstrings that don't conform to that convention are brought into conformance as part of this work.
3. Private (`_`-prefixed) symbols are exempt from the docstring-presence requirement. The private helpers without docstrings stay as they are.

### Preserved behaviour

The product's behaviour and generated output are unchanged. uncoded's index does not capture docstrings today, so this work touches source docstrings and dev tooling only.

### System non-goals

None identified.

Closes #166


alimanfoo:
## Code Analysis

### Structural baseline

**Where a docstring check would plug in.** ruff already runs as a pre-commit hook (`.pre-commit-config.yaml:16-21`), and CI's `lint` job runs `uv run pre-commit run --all-files` (`.github/workflows/ci.yml:19`). A pydocstyle rule added to ruff's config is therefore enforced in both pre-commit and CI through one change. No new CI step is needed.

**ruff's current scope.** `[tool.ruff]` sets `src = ["src"]` and `extend-exclude = [".uncoded"]` (`pyproject.toml:59-62`), and otherwise does not restrict which files are linted, so ruff lints both `src/` and `tests/`. The `select` list is `E, F, I, UP, B, SIM` (`pyproject.toml:64-72`). The `D` (pydocstyle) family is not selected, there is no `[tool.ruff.lint.pydocstyle]` section, and there are no `per-file-ignores`.

**uncoded's roots are separate.** `[tool.uncoded] source-roots = ["src", "tests"]` (`pyproject.toml:42-44`) drives uncoded's own index, not ruff. The two are independent.

**Current docstring convention.** Source docstrings are plain prose: a one-line summary, sometimes a blank line then more prose (for example `src/uncoded/body.py:9-15`). No file uses formal `Parameters`/`Returns`/`Args` section headers. Every public symbol in `src/` already has a docstring. Eleven private helpers in `src/` have none. Nothing enforces any of this today; it holds by convention only.

### What a pep257 `D` check finds today

**`src/` is already almost conformant.** Selecting `D` with `convention = "pep257"` over `src/` yields two findings: `src/uncoded/instruction_files.py:53` (D301, use `r"""` because the docstring contains backslashes) and `src/uncoded/resolver.py:31` (D105, missing docstring on the magic method `NamePath.__str__`).

**`tests/` is the opposite.** The same check over `tests/` yields 443 findings, almost all missing-docstring rules: D102 (missing method docstring) ×380, D101 (missing class docstring) ×49, D100 (missing module docstring) ×11, plus D103 ×2 and D104 ×1. Test methods are public symbols with self-documenting names, so whether the convention applies to `tests/` or to `src/` only is the central scoping question. ruff's `per-file-ignores` can exempt `tests/**` from `D`.

### Observations

**The rule has no home and nothing enforces it.** "Every public symbol carries a conforming docstring" is a rule each site follows by hand. Docstring problems recur across closed issues (#49, #57, #63, #72, #100, #110), which is the evidence that a check is warranted.

**Presence and format are not what recurred.** Those historical issues are about docstring content drifting from the code it describes. A pydocstyle `D` check verifies a docstring is present and well-formed, not that its content is accurate. The check serves consistent presence and style, but does not catch content drift.

**Edge cases.** The two `src/` findings (D105 on a magic method, D301 on a backslash-bearing docstring) are explicit choices: require docstrings on magic methods or ignore D105, and switch the one docstring to a raw string or rephrase it.

alimanfoo:
## Session Scope

This scope resolves one ambiguity in the requirements. The requirements say "every public Python symbol." Test functions, classes, and modules are technically public symbols, but their names already document intent and a docstring on each defends nothing. So this scope reads "public symbol" as the source surface (`src/`), not the test suite.

1. A standing docstring check enforces the pep257 convention over `src/`, wired into the existing ruff / pre-commit / CI path so a violation fails both pre-commit and CI.
2. `tests/` is exempt from the docstring requirement. This refines the literal "every public symbol" wording, on the grounds that test docstrings defend no behaviour.
3. `src/` docstrings are brought into conformance with the convention. Two findings exist today (`instruction_files.py:53`, `resolver.py:31`); magic-method and backslash handling are decided in Design.

alimanfoo:
## Design

Configure ruff's pydocstyle (`D`) rules, since ruff already runs in pre-commit and CI. No new tool, hook, or CI step.

Changes to `pyproject.toml`:

- Add `"D"` to `[tool.ruff.lint]` `select`.
- Add `[tool.ruff.lint.pydocstyle]` with `convention = "pep257"`.
- Add `[tool.ruff.lint.per-file-ignores]` with `"tests/**" = ["D"]`, exempting the test suite.

Bring `src/` into conformance with the two findings the convention surfaces:

- `instruction_files.py`: change the `_apply_section` docstring to a raw string (`r"""..."""`). It describes `\r\n` with literal backslashes, so a raw string is both the D301 fix and the clearer form.
- `resolver.py`: add a one-line docstring to `NamePath.__str__` (D105).

The convention keeps D105 active, so the rule applies uniformly to every public symbol including magic methods, with no config carve-out. The config change and the two fixes land together, so the hook is green from the first commit.

### Alternatives considered

- **Standalone `pydocstyle` as a separate pre-commit hook.** Delivers the same scope, but adds a second tool, dependency, and hook doing what ruff's `D` already does, for no added coverage.
- **Set aside:** `interrogate` (measures docstring coverage, not pep257 style); a hand-rolled check script; a hand-picked explicit `D`-code list (more config to own for marginal legibility).

alimanfoo:
## Plan

One task, because the config change and the source fixes must land in one commit or the ruff hook fails.

**Task 1 — Enable the pep257 docstring check in ruff and bring `src/` into conformance.** In `pyproject.toml`: add `"D"` to `[tool.ruff.lint]` `select`, add `[tool.ruff.lint.pydocstyle]` with `convention = "pep257"`, and add `[tool.ruff.lint.per-file-ignores]` with `"tests/**" = ["D"]`. Bring `src/` clean under the new rules (zero `D` findings in `src/`): today the backslash docstring in `instruction_files.py` becomes a raw string and `NamePath.__str__` in `resolver.py` gains a one-line docstring. Confirm `tests/` is exempt and the full pre-commit run is green.

alimanfoo:
## Review — fresh eyes

**Recommendation:** looks good, no blocking concerns — clean to merge.

This is a tightly scoped maintenance change. It enables ruff's pydocstyle (`D`) rules under the pep257 convention and conforms the two spots that needed it. My cold read of the diff matched the PR's stated intent without help, and I confirmed the enforcement and behaviour claims hold.

What I verified:

1. **Behaviour preserved in the docstring conversion.** In `instruction_files.py`, `_apply_section` was switched to a raw docstring (`r"""`) and `\\r\\n` rewritten to `\r\n`. A normal string renders `\\r\\n` as `\r\n`; a raw string renders `\r\n` as `\r\n`. The displayed docstring text is byte-identical, so this is a pure conformance fix (D301), not a wording change.

2. **Enforcement is real in both places.** Neither pre-commit config nor CI was touched, but both already run ruff: `.pre-commit-config.yaml` has the `ruff` hook, and the CI `lint` job runs `pre-commit run --all-files`. Adding `D` to the pyproject `select` flows to both automatically. The PR's "fails in pre-commit and CI" claim holds without extra wiring.

3. **Tree-wide conformance.** The diff only adds one docstring (`NamePath.__str__`, a magic method that D105 requires) yet enables a tree-wide rule. I checked the other 15 `src/` modules: every public module, class, function, and method already carries a docstring, no hidden `# noqa: D` suppressions exist, and the only backslash-bearing docstring is the one this PR fixed. Private (`_`-prefixed) symbols without docstrings are exempt by pydocstyle's public/private model, which matches improvement goal 3 exactly. The generated `.uncoded` stubs are excluded via `extend-exclude`, so they don't trip the rule.

### Readability

Nothing blocking. The one spot a cold read could not settle from the diff alone was tree-wide conformance: a 3-file diff cannot show that the other 15 modules already pass. That is inherent to any lint-enabling change, not a defect in this one, noted only to be transparent about where the read went beyond the diff.

No non-blocking findings, nits, or out-of-scope items.

alimanfoo:

**Recommendation:** ready to merge. No findings.

**Completeness against requirements and scope**

All three scope items are delivered.

The check is in place: `"D"` added to ruff's `select` list with `convention = "pep257"` under `[tool.ruff.lint.pydocstyle]`. The existing ruff pre-commit hook picks it up automatically, and CI runs `pre-commit run --all-files`, so the check fails on violation in both places without any new hook or CI step.

The `tests/` exemption is in place: `"tests/**" = ["D"]` under `[tool.ruff.lint.per-file-ignores]`.

`src/` is clean: the D301 fix raw-strings the `_apply_section` docstring (same text, just written as `r"""..."""` so the `\r\n` references are unambiguous), and the D105 fix adds a one-line docstring to `NamePath.__str__`. Running `ruff check src --select D` returns zero findings.

**Coherence across the whole diff**

The diff is three files. Each change is self-contained. Nothing was dropped: the `_apply_section` docstring text is preserved byte-for-byte after the raw-string conversion. The new `NamePath.__str__` docstring matches the implementation exactly. The pep257 convention shorthand disables the conflicting D2xx rules automatically, so no hand-picked ignore list is needed.

`NamePath.__str__` is the only explicitly defined dunder method in `src/`, confirmed from the namespace map, so D105 is fully resolved with no missed instances.

**Out of scope but noticed**

None.

alimanfoo:
## Response to reviews

Both reviews recommend merge with no findings, so no response work was needed. The two reviewers independently confirmed the change delivers all three scope items, that enforcement flows through the existing ruff hook and CI with no new wiring, that the docstring conversion preserves the text byte-for-byte, and that `src/` is the only tree touched (tests exempt, generated stubs excluded, private symbols exempt).

Issue #80, opened 2026-05-01
[claude] Source IO uses bare read_text/write_text without explicit encoding


alimanfoo:
Eight call sites under `src/` use `Path.read_text()` / `Path.write_text()` without an explicit `encoding=` argument, leaving the codec to interpreter / platform defaults:

- `src/uncoded/sync.py:27, 30, 33`
- `src/uncoded/instruction_files.py:169`
- `src/uncoded/serena_setup.py:131, 141, 156, 166, 189`
- `src/uncoded/extract.py:120` (in `iter_source_files`)

Python defaults to UTF-8 since PEP 540 (Python 3.7), so this is not a current bug on a stock Python installation. The concern is that the project's IO behaviour is sensitive to interpreter and platform defaults — a future Python version that changes default encoding policy, or a runtime environment where `LC_ALL` or `PYTHONIOENCODING` overrides defaults, would silently produce byte-misaligned reads or writes.

The most session-adjacent site is `extract.py:120`. PR #76's #74 fix added `ast.parse` immediately after the `read_text()` there. A source file with a PEP 263 encoding declaration would be read under the wrong codec before reaching the parser; the parser would see whatever bytes happened to land.

Possible direction: pass `encoding="utf-8"` to every `read_text` / `write_text` call. Insulates against locale and policy drift; makes the project's IO contract explicit at every site.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

alimanfoo:
Closing as won't fix. It's defensive against a Python encoding-policy change that hasn't happened. Don't add error handling, fallbacks, or validation for scenarios that can't happen.

PR #189, opened 2026-06-23
Maintenance cleanup: file-read encoding, config-error signalling, skill provenance, contributor docs

### LINKED ISSUE #178: File-read encoding handling is inconsistent and partly unguarded

### LINKED ISSUE #180: _sync_code_artefacts signals config errors with a fragile None sentinel
Noticed during #177. `_sync_code_artefacts` (`src/uncoded/cli.py`) returns `int | None`, where `None` means a validation error occurred and the caller must `return 1`. A future caller that forgets the `if result is None` check would silently swallow validation errors.

Suggested direction: raise `ConfigError` (already defined in `src/uncoded/config.py`) for the validation failures instead of returning `None`, so the error path is enforced by the type rather than a caller-side convention. `_sync` already handles `ConfigError` at its boundary.

### LINKED ISSUE #185: Generated SKILL.md files carry no "generated — do not edit" provenance
## Problem

The generated skill files under `.claude/skills/<name>/SKILL.md` and `.agents/skills/<name>/SKILL.md` give no signal that they are generated or where their source lives. They open straight with YAML frontmatter and the body:

```
---
name: uncoded-code-navigation
description: ...
---

# Code Navigation
...
```

Contrast the other generated artefacts, which announce themselves:

```
# Symbol index of this codebase, for agent navigation.
# Generated by uncoded — do not edit; regeneration overwrites.
```

(`.uncoded/namespace.yaml`, `.uncoded/docs.yaml`).

## Why it matters

During review of #184, review suggestions were posted directly on the generated `.agents/skills/.../SKILL.md` files twice, because nothing in those files signalled they were generated or pointed to the real source. The edits actually belong in `src/uncoded/skill.py` (the `SKILLS` registry, for `description`) and the body files `src/uncoded/code_navigation.md` / `doc_navigation.md` / `coherence_review.md`.

`uncoded check` already catches *committed* drift in CI, so the repo is protected against a stray edit landing — but only after the fact. A provenance marker would prevent the wasted round-trip at authoring/review time.

## Constraint

A leading `#` comment cannot be prepended before YAML frontmatter — it would break frontmatter parsing. The marker needs another home, e.g.:

- a frontmatter field (e.g. a comment-style key the framework ignores), or
- a line / HTML comment (`<!-- generated by uncoded; edit src/uncoded/... instead -->`) immediately after the closing `---`, or
- a first body line under the H1.

Whichever is chosen should name the source file so a reviewer knows where to comment.

### LINKED ISSUE #188: Contributor docs omit the docstring policy and the dev-dependency setup step
The repo expects two things of a contributor that it never states in a place they meet before CI, so a newcomer (human or agent) discovers each only when a check fails.

First, the docstring policy and its rationale live only in `pyproject.toml`. The `"D"` entry in `[tool.ruff.lint]` `select`, `convention = "pep257"` under `[tool.ruff.lint.pydocstyle]`, and `"tests/**" = ["D"]` under `per-file-ignores` together encode a policy: public symbols need a pep257 plain-prose docstring, magic methods included, private (`_`-prefixed) symbols exempt, tests exempt. Nothing says plain prose was chosen deliberately over numpy or google style. A reader can't tell from the config why a change touching docstrings adds so few of them, or which style to follow when adding one.

Second, `CLAUDE.md`'s Commands section lists `uv run pytest` and `uv run pre-commit run --all-files` but not `uv sync --extra dev`, which has to run first or the `ty` hook fails on missing dev dependencies. The README's Dev setup section has the install step; the command list a contributor actually reaches for does not.

Suggested direction: state the docstring policy and the reason for plain prose in a short contributor note (a comment by the pydocstyle config block, or a line in `CLAUDE.md`/a `CONTRIBUTING.md`), and add `uv sync --extra dev` to the `CLAUDE.md` Commands section.

alimanfoo:
This maintenance session addresses four cleanup issues surfaced after #177: file-read encoding handling (#178), config-error signalling (#180), provenance markers on generated skill files (#185), and contributor docs (#188).

## Requirements Analysis

Session Type: maintenance. Four independent improvement concerns on already-correct behaviour. Each goal is stated as a checkable property.

### #178 — file-read encoding handling

- Honour declared source encoding: uncoded reads a Python source file under its declared PEP 263 encoding, with UTF-8 as the fallback, so a validly-declared non-UTF-8 file (for example `# -*- coding: latin-1 -*-`) is indexed rather than skipped. Covers the source readers behind `sync`, `body`, and `refs`.
- No unguarded read: every file read uncoded performs handles a corrupt, unreadable, or mis-encoded file with a clean error and exit 1, or a graceful skip, never an unhandled traceback.
- Consistent handling: reads and writes of the same kind use the same mechanism, rather than a mix of bare `read_text()` and explicit UTF-8.
- Note: #178's first bullet (`sync_instruction_file`) is moot. #184 removed the instruction-file injection mechanism, so that site is gone.

### #180 — config-error signalling

- `_sync_code_artefacts` signals a validation failure by raising `ConfigError` (handled at the `_sync` boundary, as `read_config` already is) rather than returning a `None` sentinel a caller must remember to check. The return type no longer carries a `None`-means-error case.

### #185 — SKILL.md provenance

- Every generated `SKILL.md` carries a visible "generated by uncoded — do not edit; regeneration overwrites" marker, matching the marker `namespace.yaml` and `docs.yaml` already carry. It does not name an internal source path.

### #188 — contributor docs

- The docstring policy a contributor must follow (public symbols need a pep257 plain-prose docstring, magic methods included, private and test symbols exempt) is stated where a contributor meets it before CI fails, not only encoded in `pyproject.toml`.
- The `uv sync --extra dev` step appears in the `CLAUDE.md`/`AGENTS.md` Commands section, not only in the README.

### Preserved behaviour

- Index artefact content and format are unchanged for well-formed UTF-8 inputs.
- Markdown doc reads and uncoded's reads of its own artefacts stay UTF-8 (no PEP 263 mechanism applies).
- The CLI's external contract on a config error (clean message, exit 1) is unchanged. #180 is an internal mechanism change.
- Generated `SKILL.md` frontmatter still parses, and skill content is otherwise unchanged.
- `uncoded check` still detects committed drift in CI.

### Non-goals

None recorded.

Closes #178
Closes #180
Closes #185
Closes #188


alimanfoo:
## Code Analysis

### File-read layer (#178)

uncoded reads source and doc files in two different ways.

- Guarded UTF-8 helper: `_read_file_text_as_utf8` (`read_helpers.py:7`) reads UTF-8 and returns `None` on `UnicodeDecodeError`/`OSError`, printing a `warning: skipping ...` line. The sync-path source reader `iter_source_files` (`extract.py:97`) and the doc reader `iter_doc_files` (`docs_map.py:112,118`) route through it. This is the skip-and-warn convention #177 established for the sync iterators. No check enforces routing through it; it holds by convention.
- Unguarded bare reads: the `body` and `refs` commands read source with a bare `read_text()` and no encoding argument, at `body.py:16`, `resolver.py:55` (`resolve_ast_node`, used by `refs`), and `refs.py:160`. The CLI handlers `_body`/`_refs` (`cli.py:201-254`) catch `FileNotFoundError` and `SyntaxError` but not `UnicodeDecodeError`. So `uncoded body` or `uncoded refs` on a non-UTF-8 file raises an unhandled traceback, not a clean `Error:`/exit 1. The docstrings at `body.py:10` and `resolver.py:48` list the exceptions they let propagate and omit `UnicodeDecodeError`.
- Artefact reads/writes: `sync.py` reads its own prior output to compare (`sync.py:41`) and writes artefacts (`sync.py:38,44`) with bare calls, no explicit encoding.

So "read a Python source file" has split semantics: the sync path forces UTF-8 and skips a non-UTF-8 file, while `body`/`refs` read under the platform-default codec and crash on bad bytes. Nothing in the codebase uses `tokenize.detect_encoding`, so a file declaring a non-UTF-8 PEP 263 encoding is never read under that encoding anywhere.

### Config-error signalling (#180)

`_sync_code_artefacts` (`cli.py:19-83`) returns `int | None`. It returns `None` at two validation failures, a source root outside the project (`cli.py:53`) and a source root that is not a directory (`cli.py:60`), printing the error itself. The caller `_sync` translates the sentinel: `if code_result is None: return 1` (`cli.py:143`).

The doc-root validation inside `_sync` (`cli.py:154-174`) does the same two checks inline, printing and `return 1` directly with no sentinel. So the same validation concern has two shapes.

`ConfigError` (`config.py:8`) already exists. `read_config` raises it for ambiguous config, and `_sync` catches it at its boundary (`cli.py:110-112`). The type-enforced error channel is already in place; `_sync_code_artefacts` does not use it.

### Generated-artefact markers (#185)

`namespace.yaml` and `docs.yaml` each open with a two-line marker built from a `HEADER` constant prepended at render (`namespace_map.py:8`, `docs_map.py:11`): "Generated by uncoded — do not edit; regeneration overwrites."

`SKILL.md` files carry no such marker. `_render_content` (`skill.py:71-81`) renders `---\n{front}---\n\n{body}`: YAML frontmatter then body, nothing else. The render wrap is the single home for all three skills' output, so a marker added there reaches every `SKILL.md` without touching the body files. A leading `#` comment before the frontmatter would break YAML parsing. The repo's markdownlint config disables `MD041` and `MD033`, so a marker line or HTML comment after the closing `---` passes the linter.

### Contributor docs (#188)

The docstring policy lives only in `pyproject.toml`: `"D"` in `[tool.ruff.lint]` select (`pyproject.toml:66`), `convention = "pep257"` (`pyproject.toml:76`), and `"tests/**" = ["D"]` in per-file-ignores (`pyproject.toml:79`).

`AGENTS.md` (and `CLAUDE.md`, a symlink to it) has a Commands section listing the `uvx uncoded` subcommands and `uv run pytest`, but not `uv sync --extra dev`. The README's Dev setup section carries that step (`README.md:246`). The `ty` pre-commit hook needs the dev dependencies, so a contributor who skips `uv sync --extra dev` hits a hook failure with no command-list pointer to the fix.

### Code smells

1. Split read semantics, no enforced home (#178): two ways to read a source file, differing in encoding and error handling, with no check keeping reads on one path. The `body`/`refs` path can surface an unhandled `UnicodeDecodeError`.
2. `None` sentinel for an error path (#180): a caller-side convention the type does not enforce, with a parallel inline doc-root validation sibling using a different shape.
3. Inconsistent provenance markers (#185): two of three generated-artefact families carry the marker; `SKILL.md` does not.
4. One fact, not where the reader meets it (#188): the docstring policy sits only in machine config; the dev-setup step sits only in the README, not the contributor command list.

alimanfoo:
## Session Scope

Coherent Scope (accepted). Items are outcomes the work must achieve; the how is Design's call.

### #178 — file-read encoding

- a1. Every Python source read (behind `sync`, `body`, `refs`) reads the file under its declared PEP 263 encoding, UTF-8 default, so a valid non-UTF-8 file is read rather than skipped or crashed.
- a2. No source read surfaces an unhandled `UnicodeDecodeError`/`OSError`: `body`/`refs` give a clean `Error:`/exit 1, `sync` keeps skip-and-warn.
- a3. Source reads route through one mechanism (a single home for "read a source file", no stray bare `read_text`). Reads and writes of uncoded's own artefacts and of Markdown docs use explicit, consistent UTF-8.
- a4. The "no bare `read_text`/`write_text`" rule is enforced by an automated check, so it cannot silently drift again. The bare-read surface has recurred across #80, #169, #177, and #178; enforcing the rule is the root-cause fix.

### #180 — config-error signalling

- b1. `_sync_code_artefacts` signals validation failures by raising `ConfigError`, caught at the `_sync` boundary; no `None` sentinel.
- b2. The parallel doc-root validation in `_sync` uses the same `ConfigError` shape, so source-root and doc-root validation read as one model.

### #185 — provenance marker

- c1. Every generated `SKILL.md` carries a visible "generated by uncoded — do not edit; regeneration overwrites" marker, with the frontmatter still parsing.
- c2. The provenance sentence has one home: the three render paths derive the marker text from one source, not three hand-written copies.

### #188 — contributor docs

- d1. The docstring policy (public symbols need a pep257 plain-prose docstring, magic methods included, private and test symbols exempt) is stated where a contributor meets it before CI fails.
- d2. `uv sync --extra dev` appears in the `AGENTS.md`/`CLAUDE.md` Commands section.

alimanfoo:
## Design

### #178 — reads

- New `read_source_text(path) -> str` in `read_helpers.py`, using `tokenize.open` (native PEP 263 detection, BOM, UTF-8 default). Every `.py` reader uses it: `iter_source_files` (`extract.py`), `resolve_ast_node` (`resolver.py`), `resolve_body` (`body.py`), and the `refs` document read (`refs.py`).
- Error policy is carried by which function the caller uses, not a flag. The `sync` path uses a skip-and-warn wrapper `read_source_text_or_warn(path, *, warning_path) -> str | None`, preserving today's behaviour, so a valid latin-1 file is now read rather than skipped. `body`/`refs` call `read_source_text` directly and let it raise; the CLI `_body`/`_refs` handlers gain the read-failure exception types and print a clean `Error:`/exit 1.
- Rename `_read_file_text_as_utf8` to `read_doc_text_or_warn`, dropping the underscore now it sits beside the public read helpers (param `warning_path`). Still UTF-8, used by the Markdown doc reader.
- `sync.py` artefact reads and writes get explicit `encoding="utf-8"`.
- a4: an AST-based pytest test scans every `.py` under `src/uncoded` and fails if any `read_text`/`write_text` call lacks an `encoding=` keyword. No new dependency, runs in CI, reliable.

### #180 — config errors

- `_sync_code_artefacts` raises `ConfigError` at its two validation failures, so its return type becomes `int` with no `None`. The doc-root validation block raises `ConfigError` too. One `try/except ConfigError` boundary in `_sync` catches all root-validation failures, prints `Error:`, and returns 1. The `if code_result is None` branch is removed.

### #185 — provenance marker

- A dedicated small module holds one constant, `GENERATED_MARKER = "Generated by uncoded — do not edit; regeneration overwrites."`. `HEADER` and `DOCS_HEADER` compose it as a `# ` comment line; `_render_content` wraps it as `<!-- {GENERATED_MARKER} -->` placed after the closing frontmatter `---`, before the H1. It lints clean (MD033 and MD041 are disabled) and renders invisibly. Regenerating the index updates the repo's own committed `SKILL.md` files.

### #188 — contributor docs

- Add `uv sync --extra dev` as the first dev command in the `AGENTS.md` Commands section (`CLAUDE.md` is a symlink to it). Add a short "Docstrings" note stating the policy: public symbols, magic methods included, need a pep257 plain-prose docstring; private and test symbols are exempt. State it once in `AGENTS.md`, with a one-line pointer comment at the pydocstyle config in `pyproject.toml`.

### Alternatives considered

- Marker constant in `sync.py`: rejected, it adds an import edge from the content generators to the write layer that `cli.py` currently mediates.
- a4 as a pre-commit grep hook: rejected, string matching is less reliable than the AST scan.
- A visible blockquote `SKILL.md` marker: rejected, noisier in rendered output than an invisible HTML comment.

alimanfoo:
## Plan

Eight tasks, one commit each. T5 lands only after T1–T4 remove the bare reads.

- T1 — Route every source read through a declared-encoding helper. Add `read_source_text` (`tokenize.open`) and `read_source_text_or_warn(path, *, warning_path)` to `read_helpers.py`; route `extract.iter_source_files` (skip-and-warn), `resolver.resolve_ast_node`, `body.resolve_body`, and the `refs` document read through them. Same commit updates the adjacent surfaces: the `test_skips_non_utf8_file_with_warning` name and comment, the `read_helpers.py` module docstring, and the exception-enumerating docstrings on `resolve_body`/`resolve_ast_node`.
- T2 — Surface source-read failures cleanly in `body`/`refs`. The CLI `_body`/`_refs` handlers cover the read-failure types and print a clean `Error:`/exit 1.
- T3 — Rename the doc reader: `_read_file_text_as_utf8` to `read_doc_text_or_warn` (param `warning_path`), updating the `docs_map` callers. Behaviour-preserving.
- T4 — Explicit `encoding="utf-8"` on the artefact reads and writes in `sync.py`. Behaviour-preserving.
- T5 — Enforce no bare `read_text`/`write_text` in `src`. An AST-based pytest test that fails on any such call lacking an `encoding` keyword. Green after T1–T4.
- T6 — Unify config-error signalling to `ConfigError`: `_sync_code_artefacts` and the doc-root validation both raise it, one boundary catch in `_sync`, the `None` branch removed. Behaviour-preserving externally.
- T7 — Single-source the provenance marker and add it to `SKILL.md`: a dedicated `GENERATED_MARKER` module, composed into `HEADER`/`DOCS_HEADER` and into `_render_content` as an HTML comment, with the committed artefacts regenerated.
- T8 — Contributor docs: `uv sync --extra dev` in the `AGENTS.md` Commands section, a short Docstrings policy note, and a one-line pointer comment at the pydocstyle config.

alimanfoo:
## Review: completeness and coherence against requirements and scope

Recommend merge. No findings.

### Completeness against requirements and scope

All eight scope items are delivered.

**#178 — encoding (a1–a4)**

a1: Every `.py` source read uses `tokenize.open` via `read_source_text`, covering `iter_source_files`, `resolve_ast_node`, `resolve_body`, and the refs LSP document read. A validly declared non-UTF-8 file is read, not skipped or crashed.

a2: `_body` and `_refs` catch `(OSError, UnicodeDecodeError, LookupError, SyntaxError)` alongside the existing `FileNotFoundError`. Both print a clean `Error:` line and exit 1; the sync path retains skip-and-warn.

a3: All source reads go through `read_source_text` / `read_source_text_or_warn`. Artefact reads/writes in `sync_file` carry `encoding="utf-8"`. Markdown doc reads use `read_doc_text_or_warn` with explicit UTF-8. No bare `read_text`/`write_text` remains in `src/uncoded`.

a4: `PLW1514` in the ruff select list enforces the rule on type-inferred `Path` receivers. All annotated Path parameters in `src/uncoded` are covered.

**#180 — config errors (b1–b2)**

b1: `_sync_code_artefacts` return type is now `int`. Both source-root validation failures raise `ConfigError`. The `if code_result is None` branch is removed.

b2: All four doc-root validation failures also raise `ConfigError`. One `try/except ConfigError` boundary in `_sync` catches all root-validation failures with the same handler. Both validation paths read as one model.

**#185 — provenance marker (c1–c2)**

c1: `_render_content` places `<!-- {GENERATED_MARKER} -->` after the closing frontmatter `---`, before the H1. All six committed SKILL.md files carry it. Tests verify presence and position.

c2: `GENERATED_MARKER` lives in `markers.py`. `HEADER` and `DOCS_HEADER` compose it; `_render_content` uses it directly. The YAML and Markdown outputs are each rendered from the one sentence.

**#188 — contributor docs (d1–d2)**

d1: A `## Docstrings` section in AGENTS.md states the policy in three sentences. CLAUDE.md follows via symlink. The pydocstyle config in pyproject.toml points to that section.

d2: `uv sync --extra dev` is the first command in the Commands block.

The exception-propagation chain is consistent end to end: `read_source_text` raises, `resolve_ast_node`/`resolve_body` let it propagate (docstrings updated), `resolve_name_position` and `find_refs` forward the same set (docstrings updated), and the CLI handlers catch and print. No gap in the chain.

The read-helper surface in `read_helpers.py` now has three public functions with parallel naming and matching docstring style. The old `_read_file_text_as_utf8` is gone; no stray reference remains.

The ConfigError boundary in `_sync` is structurally sound. The two separate `try/except` blocks reflect two distinct error domains; the intermediate early-exit checks between them cannot join the ConfigError boundary without changing `read_config`'s API.

The provenance-marker change is byte-identical for namespace.yaml and docs.yaml. Committed SKILL.md files match the regenerated output.

Artefact index files (stubs, namespace.yaml, docs.yaml) are current for all changed surfaces.

alimanfoo:
## Review: fresh-eyes read

**Recommendation:** approve subject to a few non-blocking items. The four strands are cleanly done and the marker is genuinely single-sourced; the items below are latent, not deliverable-breaking.

### Non-blocking

1. `tests/test_skill.py:43` and `tests/test_skill.py:290` read a generated `SKILL.md` with a bare `read_text()` (no `encoding=`). This PR adds the em-dash marker `<!-- Generated by uncoded — do not edit; regeneration overwrites. -->` to every `SKILL.md` (`skill.py:82`, rendered at line 6 of each file). On a non-UTF-8 locale, for example a Windows contributor on cp1252, these two reads now raise `UnicodeDecodeError` and fail the suite. This is the exact locale-default-read hazard #178 set out to remove, landing on a site the PR just made non-ASCII. Fix: add `encoding="utf-8"` to both reads.

2. The broadened `except (OSError, UnicodeDecodeError, LookupError, SyntaxError)` in `_refs` (`cli.py:241`) wraps the whole `find_refs` call, which reaches `_run_exchange` in `refs.py`. That function does bare dict subscripts on ty's JSON response (`response["result"]`, `loc["uri"]`, `loc["range"]["start"]["line"]` around `refs.py:194-201`). A `KeyError` from a malformed LSP response is a `LookupError` subclass, so it is caught and printed as `Error: {in_path}: {e}`. A ty protocol bug then looks like an unreadable source file, hiding the real cause. Narrowing the `LookupError` catch to only the file-read step would fix it.

3. `preview = true` in `pyproject.toml` is broader than the one rule it was added for. With category prefixes (`E`, `B`, `D`, and the rest) in `select`, ruff enables every preview-status rule in those categories, not just `PLW1514`. The tree is clean today, but a future ruff release can surface new preview rules and fail CI on code unrelated to this change. Consider selecting `PLW1514` under preview while keeping the categories on stable rules, or add a one-line comment noting the widened scope was accepted.

### Readability

1. The `read_source_text` docstring (`read_helpers.py:9-16`) says it "Raises ... LookupError if the declared codec name is unknown." `tokenize.open` does not: its `find_cookie` catches the codec `LookupError` and re-raises it as `SyntaxError` ("unknown encoding for ..."). Confirmed empirically, and `test_unknown_encoding_exits_one` passes only because `SyntaxError` is also caught. So the `LookupError` arm in all three broadened catch tuples catches nothing it was meant to, which is the root of item 2 above. Fix the docstring, and reconsider whether `LookupError` belongs in those tuples at all.

### Nits

1. `_sync` now has two identical `except ConfigError as e: print(f"Error: {e}"); return 1` blocks (`cli.py:105` and `cli.py:181`). Defensible, since the first also needs the `config is None` handling that follows it, but worth a glance to see if they can fold.

### Out of scope but noticed

1. `_body` and `_refs` (`cli.py:194-218`, `cli.py:221-250`) carry a near-identical five-arm exception handler. Pre-existing, neither introduced nor worsened by this PR. A future cleanup could single-source it.

2. The stub writer (`stubs.py`) is the one generated-artefact emitter with no provenance marker; each `.pyi` carries only its source path as the first line. #185 named `SKILL.md` specifically, so this is out of scope, but if the intent is "every generated artefact says do-not-edit," stubs are the gap.

alimanfoo:
## Response to the reviews

The completeness and coherence review raised no findings; no response work was needed there.

From the fresh-eyes review:

**Non-blocking 1 (bare `read_text` on the now-non-ASCII `SKILL.md` in tests)** — accepted and fixed in 0e28354. Added `encoding="utf-8"` to the two named reads plus a third site in `test_cli.py` that compares `SKILL.md` artefacts across run directories.

**Non-blocking 2 and Readability 1 (the `LookupError` arm)** — accepted and fixed in ef5d743. Confirmed that `tokenize.open` re-raises an unknown-codec failure as `SyntaxError`, never `LookupError`, so the arm caught nothing in the read paths and in `refs` would have swallowed a `KeyError` from a malformed LSP response. Removed `LookupError` from the three catch tuples and the five docstrings that listed it, and corrected the `read_source_text` docstring.

**Non-blocking 3 (`preview = true` wider than the one rule)** — accepted and fixed in 6d27624. Added `explicit-preview-rules = true`, so the category selectors activate only stable rules and `PLW1514` is the sole preview rule in effect.

**Nit 1 (two identical `except ConfigError` blocks)** — not folded. The two blocks guard distinct error domains, and the `config is None` and nothing-to-index early returns sit between them; merging would require changing `read_config`'s return convention, which is out of scope here.

**Out of scope 1 (`_body`/`_refs` share a near-identical exception ladder)** and **Out of scope 2 (the stub writer carries no provenance marker)** — both pre-existing and out of this session's scope. Held for post-merge triage.

PR #201, opened 2026-06-24
Follow-up cleanup from #189: encoding, provenance, docstrings, and duplication

### LINKED ISSUE #192: A UTF-8 BOM in a Markdown file drops its first heading from the doc index
Noticed during #189. A Markdown file saved with a UTF-8 byte-order mark loses its top heading in `docs.yaml`.

The doc reader `read_doc_text_or_warn` (`src/uncoded/read_helpers.py`) reads with `encoding="utf-8"`, which keeps a leading BOM. `extract_headings` (`src/uncoded/docs_map.py`) requires the `#` at column 0, so a BOM-prefixed first line is not recognised as a heading and is silently omitted.

Suggested direction: read doc files with `encoding="utf-8-sig"`, which strips a BOM when present and is a no-op otherwise. Confirm by syncing a BOM-prefixed `.md` and checking its top heading appears in `docs.yaml`.

### LINKED ISSUE #193: The explicit-encoding rule holds in src but not in the test suite
Noticed during #189. The rule that every file read and write names its encoding has been fixed repeatedly on the source side (#80, #169, #177, #178) and is now enforced there by ruff's `PLW1514`. The test suite is an unenforced gap.

`PLW1514` only fires where ruff can infer a `Path` receiver, so fixture-derived paths (`tmp_path / ...`) escape it, and many bare `read_text()`/`write_text()` calls remain in `tests/`. Some read artefacts that are now non-ASCII (`namespace.yaml`, `docs.yaml`, and the `.pyi` stubs all carry an em dash), so they are locale-sensitive on non-UTF-8 systems.

Suggested direction: bring the tests under the same rule. Either add `encoding="utf-8"` to the test IO calls, or set a UTF-8 default for the suite (for example a `PYTHONIOENCODING` setting in the pytest config), so a future non-ASCII fixture cannot reintroduce a locale-default read.

### LINKED ISSUE #195: Generated stub files carry no "do not edit" provenance marker
Noticed during #189. After #185, `namespace.yaml`, `docs.yaml`, and the generated `SKILL.md` files all carry a `Generated by uncoded — do not edit; regeneration overwrites.` marker. The `.pyi` stub files do not.

The stub renderer (`src/uncoded/stubs.py`) emits each `.pyi` with only a source-path comment as its first line, with no provenance marker.

Suggested direction: have the stub renderer prepend the shared `GENERATED_MARKER` (`src/uncoded/markers.py`) as a comment, so every generated artefact family carries the marker consistently, then regenerate the committed stubs.

### LINKED ISSUE #196: Three docstrings drifted from the code after the encoding and marker changes
Noticed during #189. Three docstrings no longer match the code this session changed.

- `_render_content` (`src/uncoded/skill.py`) says "YAML frontmatter followed by the skill body", but an HTML comment marker now sits between them.
- `iter_source_files` (`src/uncoded/extract.py`) lists "unknown encoding declaration" as a skip cause, but after the `tokenize.open` change an unknown or malformed coding cookie surfaces as `SyntaxError`, not a separate category.
- `find_refs` (`src/uncoded/refs.py`) attributes the propagated read exceptions to `resolve_name_position` only, but `read_source_text` inside the LSP path (`_run_exchange`) can raise them too.

Suggested direction: correct each docstring to match the current behaviour.

### LINKED ISSUE #197: _body and _refs duplicate a near-identical exception-handling ladder
Noticed during #189. The CLI `_body` and `_refs` handlers (`src/uncoded/cli.py`) spell out the same exception ladder inline (`UnsupportedNamePath`, `SymbolNotFound`, `FileNotFoundError`, the read-failure tuple, each with its `Error:` print), now identical after this session widened both in parallel.

Suggested direction: single-source the shared handling, for example a small helper that maps these exceptions to a clean `Error:` and exit 1, so the two commands cannot drift. `_refs` has one extra arm (`RuntimeError` for the LSP path) to preserve.

### LINKED ISSUE #198: Config-root validation is duplicated across source and doc roots
Noticed during #189. Source-root validation (`_sync_code_artefacts`) and doc-root validation (`_sync`) apply the same two checks, outside the project root and not a valid directory or `.md` file, with near-identical `ConfigError` messages (`src/uncoded/cli.py`).

A related consequence: `_sync` keeps two separate `except ConfigError` boundaries because `read_config` returns `None` for the "no config file" case rather than raising.

Suggested direction: extract a shared `validate_root` helper parameterised by root kind (and whether a lone `.md` file is allowed), single-sourcing the checks and their messages. Optionally make `read_config` raise `ConfigError` for the no-config case so a single boundary covers all validation failures.

### LINKED ISSUE #199: Nothing states that CLAUDE.md is a symlink to AGENTS.md
Noticed during #189. `CLAUDE.md` is a symlink to `AGENTS.md`, but nothing says so where an editor meets either file. A contributor, or a reviewer reading a diff that touches only `AGENTS.md` while `CLAUDE.md` also shows as modified, cannot tell whether the two have drifted or which is canonical without inspecting the filesystem.

Suggested direction: state the relationship once where an editor lands, for example a line at the top of `AGENTS.md` noting that `CLAUDE.md` is a symlink to it.

alimanfoo:
Seven follow-up issues surfaced during #189: one bug in the doc index and six coherence cleanups left behind by that session's encoding, marker, and widening changes.

## Requirements Analysis

Maintenance session. Six coherence cleanups on already-correct behaviour, plus one bug (#192), shown in its own shape below.

### #192 — UTF-8 BOM drops a file's first heading (bug)

Expected behaviour: a Markdown file saved with a UTF-8 BOM has its top heading appear in `docs.yaml`. The source is the doc-index feature, which indexes every heading.

Observed behaviour (a claim verified in analysis): the first heading is silently omitted. `read_doc_text_or_warn` reads with `encoding="utf-8"`, which keeps the BOM, and `extract_headings` requires the `#` at column 0, so the BOM-prefixed first line is not recognised as a heading.

Affected consumers: an agent orienting from `docs.yaml` loses the file's top heading, usually its H1 title and main anchor.

### #193 — explicit-encoding rule not enforced in tests

Improvement goal: every file read and write in `tests/` names its encoding, enforced so a future non-ASCII fixture cannot reintroduce a locale-default read. Today `ruff`'s `PLW1514` enforces this in `src` but misses fixture-derived paths (`tmp_path / ...`), and many bare `read_text()`/`write_text()` calls remain in tests.

Preserved behaviour: test outcomes and the existing source-side enforcement are unchanged.

This rule has been fixed repeatedly on the source side (#80, #169, #177, #178). A rule that recurs across independent sites points at a check, not a one-off patch.

### #195 — stubs carry no provenance marker

Improvement goal: every generated artefact carries the shared `GENERATED_MARKER`. The `.pyi` stubs are the one family that does not; `render_stub` emits only a source-path comment.

Preserved behaviour: stub content is otherwise unchanged and still valid; the committed stubs are regenerated to match.

### #196 — three drifted docstrings

Improvement goal: three docstrings match the code they describe again, `_render_content` (`skill.py`), `iter_source_files` (`extract.py`), and `find_refs` (`refs.py`).

Preserved behaviour: code behaviour is unchanged; docstrings only.

### #197 — `_body` and `_refs` duplicate an exception ladder

Improvement goal: the shared exception-handling ladder in `_body` and `_refs` (`cli.py`) has one home, so the two cannot drift. `_refs` keeps its extra `RuntimeError` arm for the LSP path.

Preserved behaviour: the CLI's external contract (error messages, exit 1) is unchanged.

### #198 — config-root validation duplicated

Improvement goal: the two root-validation checks (outside the project root, and not a valid directory or `.md` file) and their messages have one home, shared by source-root and doc-root validation, so the two cannot drift.

Preserved behaviour: the CLI's external contract on a config error (clean message, exit 1) is unchanged.

A related consequence is the two separate `except ConfigError` boundaries in `_sync`, because `read_config` returns `None` for the no-config case. Whether to fold that in is a design call.

### #199 — symlink relationship undocumented

Improvement goal: the `CLAUDE.md` to `AGENTS.md` symlink is stated once where an editor lands, for example a line at the top of `AGENTS.md`.

Preserved behaviour: the symlink and both files' content are unchanged.

### Non-goals

None recorded.

Closes #192
Closes #193
Closes #195
Closes #196
Closes #197
Closes #198
Closes #199


alimanfoo:
## Code Analysis

### Structural baseline

**Reading layer.** `read_helpers.py` holds three readers. `read_source_text` (line 8) uses `tokenize.open`, which detects a PEP 263 cookie or a BOM and defaults to UTF-8, so source reads are BOM-aware. `read_doc_text_or_warn` (line 44) uses `path.read_text(encoding="utf-8")`, which keeps a leading BOM. The split is deliberate: source reads honour declared encodings, doc reads are plain UTF-8. The convention "every read and write names its encoding" is enforced in `src` by ruff's `PLW1514` (a preview rule under `explicit-preview-rules`).

**Marker single-home.** `markers.py` holds `GENERATED_MARKER` as the one home for the provenance text. `namespace_map.HEADER`, `docs_map.DOCS_HEADER`, and `skill._render_content` each derive from it. `stubs.render_stub` does not.

**CLI layer.** `cli.py` orchestrates through `_sync`. Validation is split: `_sync_code_artefacts` validates source roots and raises `ConfigError` (lines 46-55); `_sync` validates doc roots inline and raises `ConfigError` (lines 152-164). `read_config` returns `None` for the no-config case rather than raising, so `_sync` carries a `None` check distinct from its two `except ConfigError` boundaries (lines 103-114 and 129-183). `_body` and `_refs` each handle errors with their own inline exception ladder.

**Docstring convention.** pep257 is enforced on `src` by ruff's `D` rules (tests ignore `D`). `D` checks a docstring's presence and format, not whether it matches the code. Docstring-versus-code accuracy holds only by convention, with no check.

### Code smells, by issue

**#192 (bug, located).** `read_doc_text_or_warn` reads with `encoding="utf-8"` (`read_helpers.py:58`), keeping any BOM. `extract_headings` skips a line unless it starts with `#` and counts `#` from index 0 (`docs_map.py:57,62`), so a BOM-prefixed first line fails the test and the top heading is dropped. The mechanism spans the two functions: the reader admits the BOM, the heading parser rejects the column it shifts. The doc path is the only reader that keeps a BOM; the source path already strips it via `tokenize.open`.

**#193 (check with a gap).** `PLW1514` is selected and applies to `tests/` (per-file-ignores drops only `D` there), but it fires only where ruff infers a `Path` receiver. Fixture-derived receivers like `(src / "foo.py").write_text(...)` and `(tmp_path / rel).read_text()` escape inference. The result: 222 bare `read_text()`/`write_text()` calls in `tests/` versus 15 with `encoding=`. The risk is concrete because `GENERATED_MARKER` carries an em dash, so `namespace.yaml`, `docs.yaml`, and the `.pyi` stubs are now non-ASCII; a bare read of them on a non-UTF-8 locale would misdecode. This is a real rule, enforced for `src`, with a hole on the test side.

**#195 (family does not derive from the home).** `render_stub` (`stubs.py`) emits `# {module.rel_path}` and a blank line as its first lines, with no `GENERATED_MARKER`. The marker has a single home, but the stub family alone does not consume it.

**#196 (three drifts, all verified).** `_render_content` (`skill.py`) says "YAML frontmatter followed by the skill body", but now inserts an HTML provenance marker between them. `iter_source_files` (`extract.py`) lists "unknown encoding declaration" as its own read-failure category, but `read_source_text` surfaces a malformed or unknown cookie as `SyntaxError`. `find_refs` (`refs.py`) attributes the propagated read exceptions to `resolve_name_position` only, but `_run_exchange` calls `read_source_text(in_path)` in the LSP path, which raises the same exceptions.

**#197 (duplicated ladder).** `_body` (`cli.py:204-215`) and `_refs` (`cli.py:232-246`) spell the same four-arm ladder inline (`UnsupportedNamePath`, `SymbolNotFound`, `FileNotFoundError`, the `(OSError, UnicodeDecodeError, SyntaxError)` tuple), each with its `Error:` print and `return 1`. `_refs` adds a fifth arm, `RuntimeError` for the LSP path. The two are identical after #189 widened both in parallel, so they can drift apart again.

**#198 (duplicated validation, plus a sentinel).** Source-root validation (`cli.py:46-55`) and doc-root validation (`cli.py:152-164`) apply the same two checks (outside the project root; not a valid directory or `.md` file) with near-identical `ConfigError` messages. Separately, `read_config` returning `None` for no-config (`config.py`) forces `_sync` to keep a `None` check distinct from its `except ConfigError` handling.

**#199 (undocumented fact).** `CLAUDE.md` is a symlink to `AGENTS.md` (confirmed on disk). Nothing at the top of either file states it, so an editor or a reviewer seeing both files in a diff cannot tell which is canonical without inspecting the filesystem.

alimanfoo:
## Scope

The committed scope for this session. Each item states an outcome; the implementation choice is settled in design.

### C1 (#192)

A UTF-8 BOM at the start of any indexed Markdown file does not affect heading extraction; the file's top heading appears in `docs.yaml`.

### C2 (#193)

Every file read and write in `tests/` names its encoding, and the rule is enforced on the test side so a new escaping or fixture-derived call cannot silently use the locale default. The rule has recurred on the source side (#80, #169, #177, #178), so a check is the root-cause fix.

### C3 (#195)

Every generated `.pyi` carries the shared provenance marker, and the committed stubs are regenerated so `uncoded check` stays green.

### C4 (#196)

Three docstrings match current behaviour: `_render_content` (`skill.py`), `iter_source_files` (`extract.py`), `find_refs` (`refs.py`).

### C5 (#197)

The exception-to-CLI-error mapping shared by `_body` and `_refs` has one home, preserving `_refs`'s `RuntimeError` arm and the external contract (messages, exit 1).

### C6 (#198)

The two root-validation checks and their messages have one home shared by source-root and doc-root validation, and `_sync` has a single config-error boundary for all config failures including the no-config case.

### C7 (#199)

The `CLAUDE.md` to `AGENTS.md` symlink relationship is stated once where an editor lands.

alimanfoo:
## Design

### C1 (#192)

Change `read_doc_text_or_warn`'s codec from `"utf-8"` to `"utf-8-sig"`, so a leading BOM is stripped at the reader and `extract_headings` is unchanged. Update the function's docstring, which the codec change shifts (it currently names "non-UTF-8 bytes" as the only warning trigger and omits the BOM-stripping behaviour).

### C2 (#193)

Annotate the bare `read_text`/`write_text` calls in `tests/` with `encoding="utf-8"`, and add a local AST-based pre-commit hook that fails on any unencoded text read or write under `tests/`. The hook reaches the fixture-derived receivers `PLW1514` cannot infer, and sits in the same slot as the existing `ty` and `uncoded` local hooks.

### C3 (#195)

In `render_stub`, prepend `# {GENERATED_MARKER}` as the first line, keeping `# {module.rel_path}` as the second, from the single `markers.GENERATED_MARKER` home. Regenerate the committed stubs and update the stub-header assertions. Add a completeness check: a test asserting every generated artefact family (namespace YAML, docs YAML, stubs, skill) carries the marker, so a new family cannot silently omit it.

### C4 (#196)

Rewrite the three docstrings to match behaviour: `_render_content` names the HTML marker between frontmatter and body; `iter_source_files` folds "unknown encoding declaration" into the `SyntaxError`/parse-failure category; `find_refs` attributes the read exceptions to both `resolve_name_position` and the LSP read path (`_run_exchange`).

### C5 (#197)

Extract `_report_lookup_error(exc, *, name_path, in_path) -> int` holding the four shared exception arms (`FileNotFoundError` checked before the `OSError` tuple). `_body` and `_refs` each catch the shared union and delegate; `_refs` keeps its separate `except RuntimeError` arm.

### C6 (#198)

Extract `_validate_root(configured, *, kind, project_root, resolved_project_root, accepts_md_file) -> Path`, raising `ConfigError` and reproducing the existing messages via `kind` and `accepts_md_file`; both the source-root and doc-root loops call it. Make `read_config` raise `ConfigError` for the no-config and nothing-to-index cases, dropping its `| None` return, and update its docstring, `_find_config_file`'s docstring, and its return annotation. `_sync` then has a single `try` and a single `except ConfigError` boundary.

### C7 (#199)

Add a one-line note under the `AGENTS.md` H1: "CLAUDE.md is a symlink to this file. Edit AGENTS.md."

### Alternatives considered

**C2 via UTF-8 mode** (`PYTHONUTF8=1`) instead of annotate plus hook: near-zero churn and reliable by construction, but the rule stays implicit, diverges from the `src` per-call convention, and depends on the mode being set in every run environment.

**C6 keeping `read_config` returning `None`** with a call-site shim before a single boundary: preserves `read_config`'s current contract but keeps a sentinel the code-shape ladder would remove.

alimanfoo:
## Plan

Ten tasks, one commit each, in order.

1. (#192) `read_doc_text_or_warn` reads with `encoding="utf-8-sig"`; update its docstring; add a BOM-prefixed Markdown regression test. `extract_headings` unchanged.
2. (#193) Add `encoding="utf-8"` to every unencoded text read or write in `tests/`. Leave the byte APIs and the already-annotated calls.
3. (#193) Add a local pre-commit hook running an AST checker that fails on any unencoded text read or write under `tests/`.
4. (#195) `render_stub` prepends the provenance marker as the first line, keeping the source-path comment as the second; regenerate the committed stubs; update the stub-header test.
5. (#195) Add a test that every generated artefact family carries the marker.
6. (#196) Fix the three drifted docstrings (`_render_content`, `iter_source_files`, `find_refs`). No code change.
7. (#197) Extract a shared exception-to-error handler; `_body` and `_refs` delegate to it; `_refs` keeps its `RuntimeError` arm. External contract preserved.
8. (#198) Extract a shared root-validation helper; the source-root and doc-root loops both call it. Messages preserved.
9. (#198) `read_config` raises `ConfigError` for the no-config case and drops its `None` return; `_sync` raises the nothing-to-index case as `ConfigError` (message unchanged) and collapses to a single error boundary; update the affected docstrings and tests.
10. (#199) Add a one-line note under the `AGENTS.md` heading recording that `CLAUDE.md` is a symlink to it.

alimanfoo:
## Review: completeness and coherence

Approve. All seven scope items are delivered and the diff is coherent.

### Completeness (C1–C7)

C1 (BOM fix): `read_helpers.py` uses `utf-8-sig`; docstring notes BOM stripping. Regression test in `test_docs_map.py` covers the bug path end-to-end.

C2 (encoding enforcement): 220 annotations across eight test files; `tools/check_test_encoding.py` added; hook registered in `.pre-commit-config.yaml`. Three unmodified test files were already clean or have no IO calls.

C3 (stub provenance marker): `render_stub` prepends the marker; all 27 committed stubs regenerated; `test_header_contains_path` updated; `tests/test_markers.py` enforces the completeness rule across all four generator families.

C4 (three docstrings): all three corrected, including the follow-on that placed the cookie `SyntaxError` in the read-failure clause of `iter_source_files` rather than the parse-failure clause.

C5 (exception ladder): `_report_lookup_error` extracted with docstring; `_body` and `_refs` delegate; `_refs` keeps its `RuntimeError` arm first.

C6 (root validation + single boundary): `_validate_root` extracted with docstring, called from both loops with the exact original messages; `read_config` drops `| None` and raises `ConfigError` for no-config; `_sync` has one `try`/`except ConfigError`; docstrings for `read_config` and `_find_config_file` updated; the dead `assert config is not None` guards removed from `test_config.py`.

C7 (symlink note): blockquote under the AGENTS.md H1, visible in CLAUDE.md via the symlink.

### Coherence

Nothing further needed. The four reworked symbols in `cli.py` (`_validate_root`, `_report_lookup_error`, and the rewritten `_body`/`_refs`) fit cleanly together. The encoding rule is enforced at the right layers: PLW1514 for `src/` and `tools/`, the new hook for `tests/`.

alimanfoo:
## Review: fresh-eyes read

Recommendation: looks good; a couple of robustness gaps in the new encoding check worth a look, otherwise clean.

I read the diff cold first. My reconstruction matched the intent on all seven issues, including the #198 design call to fold the two `except ConfigError` boundaries into one by having `read_config` raise. The two refactors preserve the CLI's external contract: error strings are byte-for-byte identical, every exception is still caught on the same arm, and the wider `try/except ConfigError` in `_sync` catches nothing it didn't catch before. The BOM fix and the marker work are both complete, with no missed sibling sites.

### Non-blocking

1. The new check misses the `Path.open(...)` method form. `tools/check_test_encoding.py` only matches a bare builtin `open(...)` (an `ast.Name`); a call like `(tmp_path / "f").open("w")` is an `ast.Attribute` with `attr == "open"` and slips through. This is the same fixture-derived pattern the check was written to catch, so a future text-mode `.open()` without `encoding=` would reintroduce the very locale-default read #193 set out to prevent. No live violation today, since no test uses `.open(`.

2. The check scans `tests/*.py` non-recursively. Any future `.py` file under a `tests/` subdirectory (a `tests/helpers/` package, say) goes unchecked, an easy structural bypass. `tests/` is flat today, so this is latent.

### Nits

1. A text-mode `open(path, "w", -1, "utf-8")` with encoding passed positionally would be flagged as missing `encoding=` and block the commit. `has_encoding` scans only keywords. Positional encoding is rare in test code, so the consequence is small.

2. In `_refs`, the `RuntimeError` arm moved from last to first in the handler chain. Harmless given the actual exception types, since none of the lookup exceptions share a hierarchy with `RuntimeError`. Worth a note only because it is a silent ordering change from the original.

### Readability

Nothing to flag. My cold read of the diff matched the stated intent on every issue, so the change explains itself well to a reader without context.

alimanfoo:
## Response to the reviews

**Completeness and coherence review:** no findings. All seven items confirmed delivered; no response work needed.

**Fresh-eyes review:**

- Non-blocking 1 (the encoding check missed `Path.open()`) and Non-blocking 2 (it scanned `tests/` non-recursively): both accepted and addressed in 765ce08. The checker now also flags text-mode `Path.open()` calls and scans `tests/` recursively, so neither call form nor a future subdirectory can bypass the rule.
- Nit 1 (positional `encoding` to `open()` would be flagged): not adopted. Passing `encoding` positionally to `open()` is rare and discouraged, and handling it would add edge-case complexity for a case the suite never uses.
- Nit 2 (the `RuntimeError` arm moved first in `_refs`): no change. The reordering is behaviourally inert, since `RuntimeError` shares no hierarchy with the lookup exceptions; both reviews confirmed this.

src/uncoded/cli.py
```diff
@@ -16,6 +16,42 @@
 from uncoded.sync import remove_file, sync_file
 
 
+def _validate_root(
+    configured: Path,
+    *,
+    kind: str,
```
alimanfoo: This could be typed as a Literal, there are only two allowed values.

src/uncoded/skill.py
```diff
@@ -73,7 +73,11 @@ def _skill_path(root: Path, name: str) -> Path:
 
 
 def _render_content(*, skill: Skill) -> str:
-    """Render the full SKILL.md content: YAML frontmatter followed by the skill body."""
+    """Render the full SKILL.md content.
+
+    Sections in order: YAML frontmatter, an HTML provenance marker, then the
+    skill body.
```
alimanfoo: This is just narrating code.

tools/check_encoding.py
alimanfoo, line 1: Is there really not another way to do this, e.g., via ruff?

AGENTS.md
```diff
@@ -1,5 +1,7 @@
 # uncoded
 
+> CLAUDE.md is a symlink to this file. Edit AGENTS.md.
```
alimanfoo:
```suggestion
CLAUDE.md is a symlink to this file. Edit AGENTS.md.
```

alimanfoo:
## Review comments addressed

- **The encoding check / "can ruff do this?"** ruff's `PLW1514` is type-based: it only fires where it can infer the receiver is a `Path`, so it misses the fixture idiom in tests and at least one src call (`skill.py`'s `Traversable` read). Enforcing the rule in src via ruff but tests via a custom check was not principled, since ruff's coverage is shape-dependent everywhere. Resolved by making the custom check the single whole-tree check: it now scans `src/`, `tests/`, and `tools/`, identifying text IO by call name rather than receiver type, and `PLW1514` (with its now-purposeless preview settings) is removed (6f3fe16). `tokenize.open` is skipped, since it auto-detects encoding. The checker was also extended to cover `Path.open()` and subdirectories (765ce08).
- **`kind` could be a `Literal`:** done, `Literal["source", "doc"]` (358888e).
- **`_render_content` docstring just narrates code:** trimmed to a one-line contract; the marker's presence stays covered by a test rather than prose (a7f5d2e).
- **Symlink note:** changed from a blockquote to a plain line (89178a9).

PR #225, opened 2026-06-25
Expand ruff check selection (phases 4-8)

alimanfoo:
Maintenance session under umbrella #207 ("Expand ruff checks"). It expands the ruff rule selection across phases 4-8 and fixes the resulting violations. uncoded's product behaviour does not change. Continues the work of #221 (phases 1-3).

Closes #211
Closes #212
Closes #213
Closes #214
Closes #215

## Requirements Analysis

Session Type: maintenance. Phases 4-8 of umbrella #207 expand the repo's own ruff rule selection and fix the resulting violations. uncoded's product behaviour does not change.

### Improvement goals (each a checkable property)

1. (#211, phase 4) ruff selects `RET`, `PLR5501`, `PLR1711`, `PLR1714`, and the tree passes. Real work: 2 `PLR1714` sites become `x in {...}`. (stated)
2. (#212, phase 5) ruff selects `PLR0911`, `PLR0912`, `PLR0913`, `PLR0915`, `PLR1702`, with `preview = true` enabled so `PLR1702` is active, and the tree passes. The counters are guardrails for future code. `PLR0911` surfaced one violation, `_check_call` in `tools/check_encoding.py` (a directory the analysis did not cover): a flat dispatch `C901` already passes, documented with a per-line noqa. The others flag nothing. (stated)
3. (#213, phase 6) ruff selects `ANN`, ignores `ANN401`, adds `ANN` to the `tests/**` per-file-ignores, and the tree passes. Real work: 3 annotations in src. (stated)
4. (#214, phase 7) ruff selects `S`, adds `S101` to the `tests/**` per-file-ignores, and the tree passes. Real work: the `ty` subprocess spawn in `refs.py` (`S603`, `S607`) and the `subprocess` import (`S404`, surfaced under preview, in `refs.py` and one test) are resolved or justified. (stated)
5. (#215, phase 8) ruff selects `PT` and `N`, and the tree passes. Real work: 2 `PT006` parametrize-name fixes and renaming the two exception classes `SymbolNotFound` and `UnsupportedNamePath` for `N818`. (stated)

### Preserved behaviour

- uncoded's product behaviour is unchanged: the index, stubs, docs map, `body`, and `refs` produce the same output. (assumed)
[...]

PR #228, opened 2026-06-25
Tidy contributor and agent guidance and pin lint/type-check tools

### LINKED ISSUE #206: AGENTS.md doesn't point to the repo's cross-cutting conventions
Across a multi-issue session the team repeatedly rediscovered conventions the repo holds but `AGENTS.md` doesn't point to, repeating the orientation work each time.

`AGENTS.md` describes the index artefacts and the `body`/`refs` commands, but not where the cross-cutting conventions live. Three came up this session: that `.uncoded/` is committed and regenerated in place here; that generated artefacts share enforced invariants (a provenance marker, and explicit-encoding reads and writes); and that these conventions sit implicitly in configs, render functions, and checks rather than anywhere a reader lands. The team reconstructed each from the code and the checks.

Possible direction: add a short orientation note that points rather than restates, so it can't drift from the code. It could record that this repo commits `.uncoded/` and keeps it current via `uncoded sync` and the pre-commit hook (a local choice; other projects may gitignore it instead); that generated artefacts share invariants enforced by checks (`test_markers`, `tools/check_encoding.py`) rather than by convention; and that cross-cutting conventions live in the code and its checks, with `namespace.yaml` the first stop. Pointers into the enforced truth, not a second home for the rules and not a "complete" enumeration of artefact families.

### LINKED ISSUE #218: The 100% branch-coverage gate and `# pragma: no branch` are undocumented
Contributors meet a 100% branch-coverage gate and run into `# pragma: no branch` with no guidance on when a pragma is acceptable.

`pyproject.toml` sets `fail_under = 100` on branch coverage, and `stubs.py` carries a `# pragma: no branch` marking a branch that is always taken. Neither the gate nor the criteria for an acceptable pragma is documented, so a contributor adding a nearby call cannot tell whether to copy the pragma, omit it, or what it guards.

Direction: a short `AGENTS.md` note on how the branch-coverage gate is met and when a coverage pragma is acceptable.

### LINKED ISSUE #220: Migrate README.md developer instructions into AGENTS.md
Developers on this repo are generally agents, so consolidate all developer instructions into AGENTS.md. README stays as end user documentation only.

### LINKED ISSUE #222: Lint and type-check tools don't run at pinned versions everywhere
Noticed during #221: the lint and type-check tools run at different versions depending on how they are invoked, so a check can pass in one place and fail in another.

Two cases. The ty pre-commit hook runs `uvx ty check src tests` with no version pin (`.pre-commit-config.yaml`), so it resolves whatever ty is current on PyPI at run time. By contrast the ty server is pinned at `TY_VERSION = "0.0.37"` in `src/uncoded/refs.py:19`, with the pre-1.0 rationale recorded in #147. And ruff is not in the dev dependencies, so an agent following the new AGENTS.md instruction to "run `ruff check --fix`" via `uvx ruff` gets an unpinned ruff that can disagree with the v0.15.19 the hook pins. A violation the unpinned binary accepts can still fail the hook.

Suggested direction: pin the ty check hook (or run it from a pinned source), and make a pinned ruff available to contributors and agents, either by adding `ruff` to the dev dependencies at the hook's revision or by pointing the AGENTS.md policy at `uv run pre-commit run ruff-check` so the hook's version is always used.

### LINKED ISSUE #224: AGENTS.md contributor setup notes invite two misreadings
Two notes in the AGENTS.md contributor and command material are accurate but easy to misread, and each tripped someone during #221.

First, the setup ordering. A contributor running pre-commit in a clean checkout can hit a spurious ty failure. The ty hook uses `language: system` in `.pre-commit-config.yaml` and reads the project venv; if `uv sync --extra dev` has not run, the venv lacks dev dependencies and ty cannot resolve imports such as pytest, so the hook fails. AGENTS.md lists `uv sync --extra dev` as required for pytest and pre-commit but does not say it must run before the first pre-commit run, or why.

Second, the uv run versus uvx note. It reads "If you are contributing to uncoded itself and need to exercise local changes, use `uv run uncoded` instead." "Need to exercise local changes" reads as "only when you have changed the uncoded source," so a doc-only change appears not to qualify. But the pre-commit `uncoded` hook runs `uv run uncoded sync` regardless of the change kind.

Suggested direction: add a sentence to the `uv sync --extra dev` block noting the ty hook reads the project venv, so the sync must precede the first pre-commit run in a clean environment; and sharpen the uv-run condition to something like "if you are working in this repo, use `uv run uncoded`."

### LINKED ISSUE #227: ruff scans tools/ too, but the repo doesn't say so
Noticed during #225. The repo does not state which directories ruff scans, so an agent enabling a new rule family can miss violations outside `src/` and `tests/`.

`ruff.src = ["src"]` in pyproject.toml controls import resolution, not scan scope, and `extend-exclude = [".uncoded"]` only names an exclusion. Ruff actually scans `src/`, `tests/`, and `tools/`. During #225 a PLR0911 violation surfaced in `tools/check_encoding.py` only after the rule was enabled, because the analysis had treated src and tests as the full surface. Stating ruff's scan scope (for example in the Linting and formatting section of AGENTS.md) would close the gap.

alimanfoo:
This maintenance session tidies uncoded's own contributor and agent guidance and the tooling config behind it. It splits README and AGENTS by audience, repairs and sharpens contributor setup notes, pins the lint and type-check tools consistently, and points AGENTS.md at the repo's cross-cutting conventions. uncoded's product behaviour does not change.

Closes #220
Closes #224
Closes #222
Closes #218
Closes #206
Closes #227

## Requirements Analysis

Session Type: maintenance. Five issues, plus #227 folded in, all tidying uncoded's own contributor and agent guidance and the tooling config behind it.

### Improvement goals (each a checkable property)

1. (#220) README.md holds only end-user documentation. No contributor or maintainer setup instructions remain in it.
2. (#220) AGENTS.md is the single home for contributor and maintainer instructions, taking over the dev setup, the `uv run` vs `uvx` policy, the dogfooding note, the Windows-symlink note, and releasing.
3. (#224) AGENTS.md states that `uv sync --extra dev` must run before the first pre-commit run in a clean checkout, and why: the ty hook reads the project venv.
4. (#224) The `uv run` vs `uvx` guidance no longer reads as conditional on changing uncoded's source. It says anyone working in this repo uses `uv run uncoded`, because `uvx uncoded` runs the published release and misleads here. It strips the provenance header and reports every stub as "would update".
5. (#222) The ty check hook and the ty server run at one single-sourced pinned version, bumped from 0.0.37 to the latest release (0.0.53).
6. (#222) A pinned ruff matching the hook's version is available to contributors and agents, so a local `ruff check` agrees with the hook and CI.
7. (#218) AGENTS.md documents the 100% branch-coverage gate and the criteria for an acceptable `# pragma: no branch`.
8. (#206) AGENTS.md points to, rather than restates, where cross-cutting conventions live and are enforced: that `.uncoded/` is committed and regenerated in place; the provenance-marker and explicit-encoding invariants enforced by `test_markers` and `tools/check_encoding.py`; the C901 ceiling and the extract-a-helper response; `namespace.yaml` as the first stop.
9. (#206) The dangling pointer in AGENTS.md, "How to read and edit code in this codebase", to a heading that no longer exists, is repaired or removed.
10. (#206) The config-to-AGENTS pointers are symmetric. `max-complexity = 8` carries a pointer to its policy home the way pydocstyle already does.
11. (#206, #227) `tools/` is documented: its role as custom check scripts complementing ruff, that ruff lints the whole tree so `tools/` is in scope, and that it sits outside the coverage gate.

### Preserved behaviour

- uncoded's product behaviour is unchanged: the index, stubs, docs map, `body`, `refs`, and `check` produce the same output.
- CI keeps gating via `pre-commit run --all-files`. The 100% branch-coverage gate, the pep257 convention, the ruff rule selection, and the per-file ignores stay, except where goals 5 and 6 add pins.
- README stays valid as the PyPI long description.
- ty stays pinned, keeping #147's pre-1.0 stability intent, now single-sourced and at 0.0.53 rather than 0.0.37.

### System non-goal

- README is not a contributor guide. It is the end-user and PyPI front page. This is the boundary the README/AGENTS split must hold.

### Note

Bumping ty from 0.0.37 to 0.0.53 may surface new type diagnostics. Resolving any to keep the tree green is part of this session.


alimanfoo:
## Code Analysis

A factual read of the surfaces this work touches, with citations.

**1. Contributor guidance is split across two audience docs.** `README.md` is the PyPI long description (`readme = "README.md"`, `pyproject.toml:11`). It mixes end-user material (what it generates, install, configure, use, `body`/`refs`/`check`, how agents use it, coherence review, upgrading) with contributor and maintainer material: "Dev setup" (`README.md:248-288`), "Note for Windows contributors" (`README.md:290-295`), and "Releasing" (`README.md:342-358`). `AGENTS.md` already carries the contributor-facing Commands, Linting and formatting, and Docstrings sections. So contributor content lives in both files.

**2. The `uv run` vs `uvx` guidance is duplicated and drift-prone.** It appears in `README.md:278-281` and again, differently worded, in `AGENTS.md` Commands. Both phrase the condition as changing uncoded's source. The misreading recurred in #225, where `uvx uncoded sync` stripped the provenance marker from every stub.

**3. The ty version is pinned in one place, the check hook in none.** `TY_VERSION = "0.0.37"` (`src/uncoded/refs.py:19`) pins the ty server that `refs` spawns (`refs.py:68`). The pre-commit ty check hook runs `uvx ty check src tests` (`.pre-commit-config.yaml:27`) with no pin, so it resolves whatever ty is current on PyPI. Two ty invocations, two versions: server pinned at 0.0.37, check floating (today 0.0.53). The pin comment cites #111 ("See #111 before bumping"), but the rationale was added under #147. The version is a Python constant, so the YAML hook config cannot import it; there is no shared home today.

**4. The ruff version is pinned only in the hook.** `ruff-pre-commit` rev `v0.15.19` (`.pre-commit-config.yaml:17`) is the sole pin. ruff is absent from the `dev` optional-dependencies (`pyproject.toml:31-37`). An agent following the "run `ruff check --fix`" instruction via `uvx ruff` gets an unpinned ruff. `preview = true` (`[tool.ruff.lint]`) sharpens this, since preview rule sets differ between ruff releases.

**5. The provenance marker is already single-sourced.** `GENERATED_MARKER` lives in `src/uncoded/markers.py:3`, imported by every render family (`stubs.py:282`, `namespace_map.py:11`, `docs_map.py:12`, `skill.py:85`). `tests/test_markers.py` asserts each family emits it. This is "one fact, one home" already done.

**6. The explicit-encoding invariant is a custom check, and `tools/` is linted but not coverage-gated.** `tools/check_encoding.py` scans `src/`, `tests/`, and `tools/` for text IO calls missing `encoding=`, replacing ruff PLW1514 which misses fixture-derived receivers (`check_encoding.py:1-12`). It runs as a pre-commit hook (`.pre-commit-config.yaml:30-34`). ruff lints the whole tree (only `.uncoded` is excluded), so `tools/` is C901-gated, but coverage is scoped to `source = ["src"]` (`[tool.coverage.run]`), so `tools/` has no coverage net.

**7. The coverage pragmas.** Four `# pragma: no branch` sit in `stubs.py` (lines 344, 408, 420, 422). Each marks a branch always taken in practice: an `if remove_file(...)` whose false arm is unreachable after prior filtering, and `if d.is_dir()` / `if not any(...)` after an `rglob` that guarantees the condition. The gate is `fail_under = 100` on branch coverage (`[tool.coverage.report]`). Nothing records when such a pragma is acceptable.

**8. The config-to-AGENTS pointer asymmetry and a dangling pointer.** `pyproject.toml:110` annotates pydocstyle with `# policy stated in "Docstrings" section of AGENTS.md`. `max-complexity = 8` (`pyproject.toml:107`) carries no pointer, though AGENTS.md's "Linting and formatting" states the C901 policy. Separately, `AGENTS.md:37-38` says "See 'How to read and edit code in this codebase' below for the dispatch rule", but no such heading exists. The dispatch rule now lives in the `uncoded-code-navigation` skill.

**Code smells.** Duplicated contributor guidance across README and AGENTS (points 1, 2), two homes for one fact that can drift. The ty version as a near-duplicated fact with no shared home (point 3), the check hook pinning nothing today. Two stale cross-references (point 8): the dangling AGENTS pointer, and the refs.py pin comment citing the wrong issue.

alimanfoo:
## Scope

The work this session commits to. Each item names the outcome, not how it is achieved.

- **S1:** README.md contains only end-user documentation. Contributor and maintainer content (dev setup, the Windows note, releasing, the `uv run` vs `uvx` note) is removed from it.
- **S2:** AGENTS.md is the single home for contributor and maintainer instructions, with no contributor-guidance duplication left between the two files.
- **S3:** the `uv run` vs `uvx` guidance reads as conditional on working in this repo, not on changing uncoded's source, and notes that `uvx uncoded` runs the published release and misleads here.
- **S4:** AGENTS.md states the clean-checkout setup ordering: `uv sync --extra dev` before the first pre-commit run, and why (the ty hook reads the project venv).
- **S5:** the ty check hook and the ty server resolve to one pinned ty version, bumped to 0.0.53. The version lives in a Python constant the YAML hook cannot import, so true single-sourcing is impossible, and the two pins are instead held in agreement by a check. Any new diagnostics from the bump are resolved so the tree stays green, and the pin comment in refs.py is corrected (it cites #111; the rationale was #147) and updated for the bump.
- **S6:** a pinned ruff matching the hook version is available to contributors and agents, so a local `ruff check` agrees with the hook and CI.
- **S7:** AGENTS.md documents the 100% branch-coverage gate and the criteria for an acceptable `# pragma: no branch`.
- **S8:** AGENTS.md points to, rather than restates, where cross-cutting conventions live and are enforced.
- **S9:** the dangling AGENTS.md pointer ("How to read and edit code in this codebase") is repaired or removed.
- **S10:** the config-to-AGENTS pointers are symmetric; `max-complexity` carries a pointer to its policy home the way pydocstyle does.
- **S11:** `tools/` is documented: its role, that ruff lints it, and that it sits outside the coverage gate. This closes #227.

alimanfoo:
## Design

Guiding call: single-source every pinned tool version (one fact, one home). No agreement checks, because there is no second copy left to drift.

**ty (S5).** The version's home stays `TY_VERSION` in `refs.py`, because `refs` is product code that pulls ty via `uvx` for end users. Bump it to 0.0.53. Add `tools/check_ty.py` (matching `check_encoding.py`'s name and shape): it imports `TY_VERSION` and runs `uvx --from ty==<TY_VERSION> ty check src tests`. The ty hook entry changes from `uvx ty check src tests` to `uv run python tools/check_ty.py`. One version string in the tree; the hook and the server both derive from it, so they cannot drift. The pin comment is corrected (the rationale is #147, not #111), updated for the bump, and widened to note the constant now governs both the LSP server and the ty-check hook. Any new diagnostics from the bump are resolved.

**ruff (S6).** Convert the ruff hook from the remote `ruff-pre-commit` (rev `v0.15.19`) to a local hook running `uv run ruff check --fix --exit-non-zero-on-fix --show-fixes` and `uv run ruff format`, backed by `ruff==0.15.19` added to dev dependencies. This matches the repo's existing local `uv run` hooks. The ruff version then has one home (the dev-dep pin), shared by the hook, CI, and a contributor's `uv run ruff check`. The preview-rule skew is gone.

**Docs.** README loses Dev setup, the Windows note, and Releasing by straight deletion, gaining a one-line Contributing pointer to AGENTS.md (S1). AGENTS.md keeps its order and migrates that content in (S2), sharpens the `uv run` vs `uvx` guidance (S3), adds the clean-checkout ordering note (S4), extends Linting for the now-pinned tools, adds a Testing section for the coverage gate and acceptable-pragma criteria (S7), a Conventions section pointing to where rules are enforced (S8) and to `tools/` (S11), and removes the dangling pointer (S9). `pyproject.toml` gains a `max-complexity` pointer comment mirroring pydocstyle (S10).

### Alternatives considered

- **Alt A (ty):** two explicit pins plus a pytest test asserting they match. Keeps two homes; a sync test detects drift rather than preventing it.
- **Alt B (ruff):** keep the remote hook, add ruff to dev deps, add an agreement test. Smaller hook change, but two homes, and it doesn't unify with the repo's local-hook pattern.
- **Alt C (AGENTS.md):** a full lifecycle reorganisation. More readable, but a much larger diff and review surface.

alimanfoo:
## Plan

1. **Single-source and bump the ty pin to 0.0.53.** Bump `TY_VERSION` in `refs.py`, add `tools/check_ty.py` that imports it and runs `ty check`, point the pre-commit hook at it, correct and widen the pin comment, and keep the tree green under the new ty.
2. **Pin ruff via a local hook backed by a dev dependency.** Add `ruff==0.15.19` to dev dependencies and convert the ruff hook from the remote `ruff-pre-commit` to a local `uv run ruff` hook, preserving the current args and file scope.
3. **Migrate contributor instructions from README to AGENTS.md.** Delete Dev setup, the Windows note, and Releasing from README (leaving a one-line Contributing pointer); land them in AGENTS.md; sharpen the `uv run` vs `uvx` guidance, including the `uvx uncoded` to `uv run uncoded` command examples; add the clean-checkout setup ordering; reflect the now-pinned tooling.
4. **Document the branch-coverage gate and acceptable-pragma criteria in AGENTS.md.**
5. **Point AGENTS.md and pyproject.toml to the repo's cross-cutting conventions.** Add the conventions pointers, document `tools/`, remove the dangling pointer, and add the `max-complexity` pointer comment mirroring pydocstyle.

alimanfoo:
## Review: completeness and coherence against scope

**Recommend merge.** All eleven scope items delivered; no coherence gaps remain.

**Completeness against scope**

Every item from the accepted scope is present in the diff.

- S1/S2: README loses all contributor content; AGENTS.md is the sole home with no duplication.
- S3: Commands intro correctly explains why `uvx uncoded` misleads; all four command examples use `uv run uncoded`. The one remaining `uvx uncoded` occurrence is the explanation of why not to use it — correct.
- S4: Dev setup states the clean-checkout ordering with the right reason (ty type-checks src/tests; test modules import dev-only packages).
- S5: `TY_VERSION` bumped to 0.0.53 in refs.py and the stub; pin comment widened; `tools/check_ty.py` imports the constant so the hook and the LSP server structurally cannot drift; the old `uvx ty check` entry is gone.
- S6: `ruff==0.15.19` in dev deps; remote ruff-pre-commit hook replaced by two local hooks; version has one home.
- S7: Testing section documents the `fail_under = 100` gate and both qualifying pragma cases; all four stubs.py instances cited and verified against the code.
- S8: Conventions section names five conventions with pointers, none restating; opening sentence describes the section without overclaiming enforcement.
- S9: Dangling "How to read and edit code in this codebase" pointer removed.
- S10: `max-complexity = 8` carries a policy pointer matching the pydocstyle comment.
- S11 (closes #227): Tools subsection names both scripts, notes ruff lints tools/, notes the coverage-gate exclusion.

**Coherence across the full diff**

No stale references remain. Cross-references within AGENTS.md resolve to sections present in the file. The pyproject.toml pointer comments name existing section headings. docs.yaml reflects the new AGENTS.md structure. README retains its integrity as an end-user/PyPI document.

No findings.

alimanfoo:
## Review: fresh-eyes read

**Recommendation:** looks good. One non-blocking fix and a small nit; no blocking concerns.

A cold read of the diff matched the stated intent closely. The README/AGENTS split, the ty and ruff single-sourcing, and the new Testing and Conventions sections all do what the description says. Two focused lenses, docs accuracy and pre-commit tooling equivalence, both came back clean: every line number, file name, symbol, and setting cited in the changed prose matches the repo, and the hook migration preserves or improves behaviour with no silent gap.

### Non-blocking

1. README.md "Contributing" section links to AGENTS.md with a relative path, `[AGENTS.md](AGENTS.md)`. PyPI does not rewrite relative links, so a reader on the project page who clicks it lands on a broken URL under pypi.org, not the file on GitHub. The README must stay valid as the PyPI long description, so use an absolute GitHub URL, for example `https://github.com/alimanfoo/uncoded/blob/main/AGENTS.md`.

### Nits

1. src/uncoded/refs.py: the "before bumping" pointer changed from `#111` to `#147`. The comment still names the reason as the `textDocument/references` edge cases, which is the history #111 recorded. Confirm the cited issue carries that edge-case context, so the next maintainer about to bump the pin reaches the evidence behind the strict pin, not just the stability-intent issue.

### Readability

1. The ty bump (0.0.37 to 0.0.53) is the only runtime-affecting line in an otherwise docs-and-config PR, because refs.py spawns the ty LSP server at `TY_VERSION`. The diff alone gives a reader no way to see that the references output is unchanged at the new version; the safety net is the integration tests, which CI runs via `pytest -m "integration or not integration"`. No change needed; flagging so whoever merges knows the bump's only runtime guarantee rests on those integration tests passing.

alimanfoo:
## Response to the reviews

Thanks to both reviewers. The completeness-and-coherence review raised no findings. The fresh-eyes review raised two fixes and one note, all handled below.

- **Contributing link (non-blocking).** Accepted. The relative link would break on PyPI, where README is the long description. Changed to an absolute GitHub URL in 6eca952.
- **ty pin pointer (nit).** Accepted, and it corrected a mistake this PR introduced. An earlier commit changed the "before bumping" pointer from #111 to #147, but #111 is the issue that carries the `textDocument/references` edge-case evidence (it lists astral-sh/ty #3406, #3387, #3369, #3411, #3410), while #147 only requested the comment. Reverted to #111 in 0cae5be.
- **ty bump runtime guarantee (readability).** No change. The bump to ty 0.0.53 is the only runtime-affecting line, since `refs.py` spawns the ty LSP server at that version. The references output's only guarantee at the new version is the integration test suite, which CI runs. Flagging it so it is visible to whoever merges.

tools/check_ty.py
```diff
@@ -0,0 +1,23 @@
...
+
+from uncoded.refs import TY_VERSION
+
+
+def main() -> int:
+    """Run the ty check at the pinned version and return ty's exit code."""
+    result = subprocess.run(  # noqa: S603 -- fixed command vector; uvx resolved from PATH by design
+        ["uvx", "--from", f"ty=={TY_VERSION}", "ty", "check", "src", "tests"],  # noqa: S607 -- uvx resolved from PATH by design
+    )
+    return result.returncode
```
alimanfoo: If ty fails, how will an agent see the failures and know how to fix them?

AGENTS.md
```diff
@@ -34,45 +34,71 @@ Index artefacts, generated from the configured roots:
...
 
-This project uses [uv](https://docs.astral.sh/uv/). Run `uncoded` commands via
-`uvx` so they run the published package without needing to install it. Run
-project tooling (`pytest`, `pre-commit`) via `uv run`. If you are contributing
-to uncoded itself and need to exercise local changes, use `uv run uncoded ...`
-instead. `uvx` always runs the published release.
+This project uses [uv](https://docs.astral.sh/uv/). Run all commands via
+`uv run`. `uvx uncoded` runs the published PyPI release, not the local editable
+install. It strips the provenance marker from generated files and reports every
+stub as "would update".
```
alimanfoo:
```suggestion
This project uses [uv](https://docs.astral.sh/uv/). Run all commands via
`uv run`. `uvx uncoded` runs the published PyPI release, not the local editable
install, which may have different behaviour.
```

AGENTS.md
```diff
@@ -34,45 +34,71 @@ Index artefacts, generated from the configured roots:
...
+`CLAUDE.md` as a plain text file containing the string `AGENTS.md` rather than
+following the symlink.
+
 ## Linting and formatting
 
-Run `ruff check --fix` and `ruff format` before committing. The pre-commit hooks
-run the same fixes automatically. If a hook rewrites files, the commit fails.
+Run `uv run ruff check --fix` and `uv run ruff format` before committing. Both
+use ruff 0.15.19, pinned via the `dev` optional dependency. The pre-commit hooks
+run the same commands automatically. If a hook rewrites files, the commit fails.
```
alimanfoo: Don't cite the ruff version here, duplicates.

AGENTS.md
```diff
@@ -85,11 +111,85 @@ reviewed the change.
 A C901 violation means a function is too complex to pass the check. Refactor or
 flatten it rather than raising the `max-complexity` threshold.
 
+The ty pre-commit hook runs the type checker, pinned at 0.0.53 via `TY_VERSION`
+in `src/uncoded/refs.py`.
```
alimanfoo: Don't cite the ty version here, duplicates.

AGENTS.md
```diff
@@ -85,11 +111,85 @@ reviewed the change.
...
+
+Use `# pragma: no branch` only when a branch's false arm is structurally
+unreachable in practice. Two cases qualify:
+
+- The branch follows an `rglob` that just yielded the item. `remove_file` on a
+  file `rglob` just found will always succeed; the false arm requires the file
+  to vanish between the two calls.
+- The condition is guaranteed true by operations immediately above it in the
+  same function. The false arm would contradict the state those operations
+  produced.
```
alimanfoo: Let's discuss, this sounds weird.

AGENTS.md
```diff
@@ -85,11 +111,85 @@ reviewed the change.
...
+
+The four uses in `stubs.py` (lines 344, 408, 420, 422) are the only examples in
+the codebase. Lines 344 and 408 are the rglob case. Lines 420 and 422 follow the
+removal of all `.pyi` files, so the directory is guaranteed empty.
```
alimanfoo: Why are we citing specific code lines here? That will just drift and is totally pointless. Come on Bridget.

AGENTS.md
```diff
@@ -85,11 +111,85 @@ reviewed the change.
...
+**Provenance marker.** Every generated file carries `GENERATED_MARKER`, defined
+in `src/uncoded/markers.py`. `tests/test_markers.py` verifies that all four
+output kinds carry it.
+
+**Explicit encoding.** Every text read/write in `src/`, `tests/`, and `tools/`
+must pass `encoding=`. `tools/check_encoding.py` enforces this as a pre-commit
+hook; it covers receiver types that ruff PLW1514 misses.
+
+**Complexity ceiling.** `max-complexity = 8` in `[tool.ruff.lint.mccabe]`. See
+"Linting and formatting" above for the response to a C901 violation.
```
alimanfoo: The actual value of `max-complexity` might change, don't duplicate it here.

pyproject.toml
```diff
@@ -104,7 +105,7 @@ ignore = [
 ]
 
 [tool.ruff.lint.mccabe]
-max-complexity = 8
+max-complexity = 8  # policy stated in "Linting and formatting" section of AGENTS.md
```
alimanfoo: Really? Do we really need a comment here? Come on Bridget.

tools/check_ty.py
alimanfoo, line 1: I am not convinced this is necessary at all. We use ty for two different things. One is as a language server, called at runtime to provide functionality. The other is as a type checker development tool. There is no need for both to be pinned to the same version. The type checker can just pin to whatever ends up in the uv lock file.

alimanfoo:
## Revised Scope and Design

Supersedes the Scope and Design comments above. During review the approach to two items changed. The rest of the Scope and Design stands.

**Scope**

- **S5 (ty pinning) — revised.** The ty type-checker (a development tool) and the ty LSP server (runtime, in `refs.py`) are independent and need not share a version. The type-checker is pinned via the `dev` dependency and the lock, the same way ruff is. The LSP server keeps its own strict pin in `refs.py`. There is no shared-version check and no wrapper. (The earlier S5 proposed holding two ty pins in agreement by a check.)
- **S10 (config pointers) — revised.** AGENTS.md is the entry point that directs to where each policy lives and is enforced. The canonical home of a policy is the file or config that enforces it, not AGENTS.md. So the config-to-AGENTS pointer comments are removed, both the `max-complexity` one this session would have added and the pre-existing pydocstyle one, rather than made symmetric.

**Design**

- **ty mechanism.** The `tools/check_ty.py` wrapper that single-sourced one shared ty version is removed. ty is a dev dependency pinned by the lock, and the hook runs `uv run ty check src tests`, matching the ruff hook. The runtime LSP server keeps `TY_VERSION` in `refs.py`, bumped to 0.0.53 to stay current.
- **Pointers.** AGENTS.md and pyproject.toml point to canonical homes and do not restate values that live in the enforcing files: tool versions, the complexity ceiling, the coverage threshold, and the pragma line numbers.

alimanfoo:
## Response to the review comments

Thanks, these sharpened the change. How each was handled:

- **ty wrapper, necessity and how failures surface.** Agreed. The type-checker and the LSP server are independent, so the `tools/check_ty.py` wrapper is removed. ty is pinned via the `dev` dependency and the lock like ruff, and the hook runs `uv run ty check`, which prints ty's diagnostics straight to the agent. (116f3c1)
- **Commands intro wording.** Applied as suggested. (9cfea6e)
- **Ruff and ty versions cited in Linting.** Removed. The versions live in the dev dependency and the lock. (116f3c1, 9cfea6e)
- **Specific `stubs.py` line numbers in Testing.** Removed the enumeration; the line numbers would drift and add nothing. The qualifying-case criteria stay. (9cfea6e)
- **`max-complexity` value duplicated.** Removed the value; the Conventions entry points to `[tool.ruff.lint.mccabe]`. The same was applied to the coverage threshold. (9cfea6e)
- **The pyproject pointer comments.** Agreed they point the wrong way. Removed both, on `max-complexity` and on pydocstyle. AGENTS.md directs to the config, which is the canonical home. (9cfea6e)
- **Prose that read awkwardly.** The affected sentences were reworked, and a copy-edit pass tidied the changed sections. (9cfea6e, 260b0e4)

The runtime ty pin was kept current at 0.0.53 per your call. The revised Scope and Design are recorded in the comment above.

alimanfoo:
## Revised Scope item S7 (pragma handling)

Supersedes S7 in the Scope above and the pragma guidance in the earlier Testing-section work. A review discussion found the documented criterion ("a `# pragma: no branch` is acceptable when the false arm is structurally unreachable") to be self-contradictory: a truly unreachable arm is dead code, so the branch should be removed, not silenced with a pragma.

Resolution: rather than document when a pragma is acceptable, the session removed all four `# pragma: no branch` from the tree.

- The two in the stub-removal loops marked an artificial branch. `remove_file` returns a bool, so `if remove_file(...): changes += 1` is just `changes += remove_file(...)`. The branch and both pragmas are gone.
- The two in the directory-pruning guards marked defensive branches whose false arm is reachable when a foreign file sits in the stubs tree. A new test plants such a file and covers both arms, so the pragmas are gone and the defensive behaviour (leave foreign files and their directories intact, do not crash) is now pinned by a test.

GH218 is still resolved. The Testing section documents the 100% branch-coverage gate, and now tells contributors to cover both arms or remove a dead branch rather than reach for a pragma. The repo carries no coverage pragmas.

AGENTS.md
```diff
@@ -85,11 +109,71 @@ reviewed the change.
...
+in `src/uncoded/markers.py`. `tests/test_markers.py` verifies that all four
+output kinds carry it.
+
+**Explicit encoding.** Every text read/write in `src/`, `tests/`, and `tools/`
+must pass `encoding=`. `tools/check_encoding.py` enforces this as a pre-commit
+hook. It covers receiver types that ruff PLW1514 misses.
+
+**Complexity ceiling.** The value is in `[tool.ruff.lint.mccabe]` in
+`pyproject.toml`. See "Linting and formatting" above for the response to a C901
+violation.
```
alimanfoo: Could link to Linting and formatting.

AGENTS.md
```diff
@@ -85,11 +109,71 @@ reviewed the change.
...
+
+**Index committed.** Commit `.uncoded/` and keep it current with the pre-commit
+hook. See "Commands" and "Dev setup" above.
```
alimanfoo: Could be links.

AGENTS.md
```diff
@@ -85,11 +109,71 @@ reviewed the change.
...
+
+**`namespace.yaml` first.** Load `.uncoded/namespace.yaml` before any symbol
+lookup. The `uncoded-code-navigation` skill has the full dispatch rule.
```
alimanfoo: This is actually redundant, the Before you start section will cover this.

AGENTS.md
```diff
@@ -85,11 +109,71 @@ reviewed the change.
...
+
+### Tools
+
+`tools/` holds custom check scripts that complement ruff. The current script is
+`check_encoding.py` (encoding invariant). Ruff lints the whole tree, so `tools/`
```
alimanfoo: Don't enumerate what tools are currently there, no point and just something else to keep in sync.

alimanfoo:
## Response to the Conventions comments

All four addressed in 9e5c89f:

- **Cross-references → links.** The Complexity ceiling and Index committed entries now use markdown links (`#linting-and-formatting`, `#commands`, `#dev-setup`) instead of "See … above".
- **`namespace.yaml` first.** Removed. It duplicated "Before you start", which already loads the navigation skill.
- **Enumerating `tools/` scripts.** Removed the "current script is …" sentence. The section keeps the role, the ruff-in-scope note, and the coverage-gate exclusion, without a list to keep in sync.

PR #230, opened 2026-06-25
Retire the custom encoding check in favour of EncodingWarning

### LINKED ISSUE #203: EncodingWarning may enforce the encoding rule without a custom script
Noticed during #201. The encoding rule (every text read and write names its encoding) is now enforced by a bespoke AST script, when CPython's own `EncodingWarning` (PEP 597) may enforce it more simply across the whole tree.

`tools/check_encoding.py` walks every `.py` file under `src/`, `tests/`, and `tools/` and statically re-derives which calls open text without an `encoding=`. The interpreter already offers this at runtime: under `-X warn_default_encoding`, any text IO that falls back to the locale default raises `EncodingWarning`, and `-W error::EncodingWarning` turns those into failures. #201 removed ruff's `PLW1514` in favour of the custom script; this issue asks whether the script is needed at all.

Suggested direction: run the test suite under `python -X warn_default_encoding -W error::EncodingWarning` and compare what it catches against the static check. It covers `open`, `Path.open`, `read_text`/`write_text`, and `io.open` natively, and `tokenize.open` does not trigger it, so the bespoke `tokenize` skip would also be unnecessary. If coverage holds, retire `tools/check_encoding.py` and its pre-commit hook. The one tradeoff to weigh: `EncodingWarning` is runtime-only, so it flags a line only when a test executes it, whereas the static check also sees unexecuted code — assess this against the suite's branch coverage.

alimanfoo:
This change addresses #203: replace the bespoke `tools/check_encoding.py` with CPython's built-in `EncodingWarning` mechanism, if the built-in covers the rule's full scope.

### Improvement goals

1. Enforce the encoding rule with CPython's built-in `EncodingWarning` mechanism (the suite run under `-X warn_default_encoding -W error::EncodingWarning`) instead of the bespoke `tools/check_encoding.py`, provided enforcement still covers the rule's full scope. Checkable: the rule is enforced, and `tools/check_encoding.py` plus its pre-commit hook are gone.
2. Every reference to the retired script is updated, notably the "Explicit encoding" convention in `AGENTS.md`, so no doc points at a deleted check.

### Preserved behaviour

The encoding rule still holds: no text read or write that omits `encoding=` can be introduced without a check failing, with one accepted residual identified in analysis and confirmed at scope. A fixture-derived receiver write (such as `(tmp_path / "f").write_text(...)`) on a test line the suite never executes is caught by neither the static rule nor the runtime gate; the old script caught it statically. `src/` is fully exercised by the coverage gate and a future tool directory is covered statically, so the residual is narrow. Product behaviour and test outcomes are unchanged; this swaps an enforcement mechanism only.

### Key consideration

`EncodingWarning` is runtime-only. It flags a line only when executed. `src/` is fully exercised by the 100% branch-coverage gate, but `tests/` is not coverage-measured, and `tools/` is not run by the suite at all. So a naive swap risks losing coverage for `tools/` and unexecuted test code. The delivered design pairs the runtime `EncodingWarning` gate with the static ruff `PLW1514` rule so the rule stays enforced across the whole tree; see the Design comment below.

### Non-goals

None recorded.

Closes #203


alimanfoo:
## Code Analysis

### Structural baseline

1. **The rule and its enforcement.** The rule (every text read/write in `src/`, `tests/`, `tools/` passes `encoding=`) has many independent call sites and no single home, so a check enforces it. The check is `tools/check_encoding.py`, wired as the `check-encoding` pre-commit hook (`.pre-commit-config.yaml:35-39`, `pass_filenames: false`). The script walks all three roots itself (`_SCAN_ROOTS`, `check_encoding.py:18`). It is documented in `AGENTS.md:138-140`.

2. **What the script detects.** It flags `open()`, `Path.open()`, `.read_text()`, `.write_text()` calls lacking `encoding=`; it skips binary mode (`"b"` in the mode literal) and `tokenize.open` (`check_encoding.py:48-62`). Name-based dispatch lets it catch fixture-derived receivers like `(tmp_path / "f.py").write_text(...)` that ruff's `PLW1514` missed. That miss was the stated reason the script replaced `PLW1514` (#193, #201).

3. **The `EncodingWarning` candidate.** Confirmed empirically: under `-X warn_default_encoding`, text-mode `open`/`read_text`/`write_text`/`io.open` calls that fall back to the locale default raise `EncodingWarning`; `-W error::EncodingWarning` makes those fail. Binary mode and `tokenize.open` do not trigger it. Its detection surface matches the script's by construction, since both target the same locale-default fallback. `requires-python = ">=3.12"` clears the 3.10 floor for the feature.

4. **Coverage scope.** `EncodingWarning` is runtime-only: it flags a line only when that line executes. The three roots differ in runtime exposure under the suite:
   - `src/`: full. The coverage gate is `source = ["src"]`, `branch = true`, `fail_under = 100`, so every line and branch executes.
   - `tests/`: partial. Test code runs but is not coverage-measured. Integration tests are excluded by default (`addopts = "--cov -m 'not integration'"`). An unencoded call on an unexecuted or skipped test path would not be seen.
   - `tools/`: none. `tools/` is never imported or run by pytest. A pytest-driven `EncodingWarning` run gives `tools/` zero coverage. Today `tools/` holds only the script itself, but the rule's scope names `tools/`, so future tool scripts would be unguarded.

5. **Where each mechanism runs.** The script lives in the CI lint job (`pre-commit run --all-files`) and local pre-commit. The suite runs in the CI test job across 3.12 and 3.13. `-X warn_default_encoding` is an interpreter start-up flag; pytest `addopts` cannot set it, so enabling it needs `PYTHONWARNDEFAULTENCODING=1` or `python -X ... -m pytest`. `-W error::EncodingWarning` can come from pytest's `filterwarnings` ini option, `-W`, or `PYTHONWARNINGS`.

### Code smells

- **Custom machinery where a built-in may serve.** The script reimplements (binary-mode parsing, `tokenize` skip, receiver matching) what the interpreter does natively.
- **Coverage mismatch between the two mechanisms.** The static script sees all code; `EncodingWarning` sees only executed code. The roots without runtime coverage (`tools/`, unexecuted `tests/`) are exactly where a swap would weaken the rule.

alimanfoo:
## Scope

Maintenance. The encoding rule is enforced today by a bespoke AST script (`tools/check_encoding.py`). The aim is to retire it for a built-in or off-the-shelf mechanism while keeping the rule enforced.

### Coherent Scope

1. The encoding rule stays enforced across all three roots (`src/`, `tests/`, `tools/`) without the bespoke AST script. Off-the-shelf tooling beyond `EncodingWarning` may be used.
2. Local pre-commit enforcement and CI enforcement are both preserved, matching today, where the script runs at pre-commit and in the CI lint job. A violation fails before merge.
3. `tools/check_encoding.py` and its `check-encoding` pre-commit hook are removed.
4. Every reference to the retired script is updated: the "Explicit encoding" convention (`AGENTS.md:138-140`), the "### Tools" section and its coverage-exclusion rationale (`AGENTS.md:149-153`, since removing the only file empties `tools/`), and any in-code comment citing `PLW1514` or the script.

Accepted residual: the realistic off-the-shelf combination leaves one narrow gap versus today, a fixture-derived-receiver text write (like `(tmp_path / "f.py").write_text(...)`) in `tests/` on a line the suite never executes, even with CI integration tests. This was the exact case the bespoke script was built to catch statically. Design minimises it; accepting this scope accepts it as the cost of dropping the script.

alimanfoo:
## Design

The encoding rule keeps full enforcement after the bespoke script is retired, using two off-the-shelf mechanisms already in the stack, guarded by a sentinel test.

### Design

Target shape when done:

1. `PLW1514` (unspecified-encoding) added to the ruff `select` list in `pyproject.toml` (preview already on). Ruff runs at pre-commit and the CI lint job, giving static enforcement across the whole tree for standard and typed-receiver calls. Zero new violations, since all such calls already carry `encoding=`. It is not added to the `tests/**` per-file-ignores, so it fires in tests too; only fixture-derived receivers fall through.
2. `filterwarnings = ["error::EncodingWarning"]` added to `[tool.pytest.ini_options]`, promoting the warning to a test failure.
3. `PYTHONWARNDEFAULTENCODING=1` set in the CI test job env, and documented for local runs. This makes text IO that omits `encoding=` emit `EncodingWarning`, catching the fixture-derived receivers in `tests/` when executed.
4. A sentinel test performs an unencoded text write and asserts `EncodingWarning` is raised. If the env var is missing, no warning fires and the sentinel goes red, converting the fail-open mode into a loud failure and making the env-var dependency self-enforcing.
5. `tools/check_encoding.py` and the `check-encoding` pre-commit hook removed. `tools/` becomes empty.
6. `AGENTS.md` updated: the "Explicit encoding" convention rewritten to name both mechanisms, the `PLW1514` blind spot, the runtime half that fills it, and the env-var dependency, with pointers to `pyproject.toml` and `ci.yml`; the whole-tree scope preserved. The "### Tools" section dropped, since `tools/` is gone.

Residual (accepted in scope): a fixture-derived receiver text write in `tests/` on a line the suite never executes.

### Alternatives considered

**A single syntactic static matcher (ast-grep or semgrep)** as a pre-commit hook, matching `open`/`read_text`/`write_text` without `encoding=` by syntax. Delivers the full scope with no residual and no fail-open mode (fully static). Not chosen: it adds a new external tool to the stack and trades the bespoke Python script for a bespoke matcher rule, where the chosen design reuses tools already present.

**`PLW1514` with the blind spot closed at source** (annotate fixture receivers so ruff resolves their type). An empirical check showed ruff does not resolve a fixture-derived receiver type through the `/` operator even when annotated `Path`, so this approach does not work and was dropped.

alimanfoo:
## Plan

**Task 1 — Enable ruff PLW1514 across the tree.** Add `PLW1514` to the `select` list under `[tool.ruff.lint]` in `pyproject.toml` (preview already on). Not added to the `tests/**` per-file-ignores. The tree is already compliant, so expect zero violations.

**Task 2 — Add the runtime EncodingWarning gate.** Add `filterwarnings = ["error::EncodingWarning"]` to `[tool.pytest.ini_options]`, and `PYTHONWARNDEFAULTENCODING=1` to the CI test job. The suite runs with that env var set. Watch-point: a third-party unencoded open under the error filter; scope the filter to first-party code if it surfaces, rather than dropping the gate.

**Task 3 — Add the sentinel test.** One test does an unencoded write and asserts `EncodingWarning` is raised; it goes red when the env var is absent, converting the fail-open mode into a loud failure.

**Task 4 — Retire the bespoke script and align the docs.** Delete `tools/check_encoding.py` and its `check-encoding` pre-commit hook. Rewrite the "Explicit encoding" convention to name both mechanisms, the PLW1514 blind spot the runtime half fills, and both silent-disable dependencies (the `PYTHONWARNDEFAULTENCODING` env var and the ruff `preview` mode), with pointers to `pyproject.toml` and `ci.yml`, keeping the whole-tree scope; drop the "### Tools" section; and update the documented test command(s) in the "Commands" section to carry `PYTHONWARNDEFAULTENCODING=1`.

alimanfoo:
## Review: fresh-eyes code review

**Recommendation:** looks good; a few non-blocking robustness and doc-clarity points, no blockers.

The change cleanly retires the bespoke `tools/check_encoding.py` and replaces it with ruff PLW1514 (static, whole tree) plus a runtime `EncodingWarning` gate. The sentinel test logic is sound: when `PYTHONWARNDEFAULTENCODING` is unset it fails with "DID NOT WARN" rather than passing green, and `pytest.warns` correctly suspends the global `error::EncodingWarning` filter inside its block, so the test passes when the var is set rather than erroring. `preview = true` is present, so PLW1514 is live. No stale references to the removed script remain.

### Non-blocking

1. **The sentinel guards the env var, but not the filter line it depends on.** `tests/test_encoding_gate.py` only asserts that `EncodingWarning` is *emitted*, which proves `PYTHONWARNDEFAULTENCODING` is set. It does not prove the warning is *escalated to an error*. If someone later deletes `filterwarnings = ["error::EncodingWarning"]` from `pyproject.toml`, real fixture-derived violations would downgrade to printed warnings, the build would stay green, and the sentinel would still pass. Half the gate would be silently off. Worth considering whether the sentinel can pin the escalation too, not just the env var.

2. **The new enforcement now rests on two unguarded config lines.** Before this PR, enforcement lived in one self-contained script. Now it hinges on `filterwarnings` (point 1) and on `preview = true` in `pyproject.toml`: PLW1514 silently drops if preview is removed, and nothing fails when it does. The `preview` dependency pre-dates this PR, but this PR makes it load-bearing for encoding enforcement for the first time. Both are documented in AGENTS.md, but documentation is not a check.

### Nits

1. AGENTS.md, "Explicit encoding" section: "Two mechanisms enforce this" states a count it then lists, which the writing-style guide advises against. Consider rewording to name the static check and runtime gate without the count.

2. AGENTS.md still names `tools/` in the encoding scope, but the `### Tools` section was removed and the directory is now empty. The lone mention reads slightly orphaned. Defensible as forward-looking, since PLW1514 covers the whole tree, but a reader may wonder why `tools/` is singled out with nothing in it.

### Readability

1. The PR goal frames the work as "enforce the encoding rule with CPython's built-in `EncodingWarning` mechanism ... instead of the bespoke script." But the delivered design makes ruff PLW1514 the primary, whole-tree static enforcement, with `EncodingWarning` as a runtime supplement that catches only the fixture-derived receivers PLW1514 cannot type-resolve. Reading the diff cold, I concluded PLW1514 does most of the work; the requirements text led me to expect `EncodingWarning` to be central.

2. The "Preserved behaviour" requirement reads as absolute: "no text read or write that omits `encoding=` can be introduced ... without a check failing." But AGENTS.md documents an accepted residual: a fixture-derived call on a never-executed test line is caught by neither mechanism. The old script caught it statically. The residual is narrow in practice, but the canonical requirement and the convention doc now state opposite things. Worth confirming the narrowing was a conscious accept, and aligning the wording if so.

alimanfoo:
## Review: completeness and coherence

**Recommend merge.** The PR delivers the agreed scope in full, and the finished change is coherent. No blocking findings.

### Completeness against the scope

1. The encoding rule stays enforced across the whole tree without the bespoke script. Static enforcement is ruff PLW1514, added to the `select` list (`pyproject.toml`) and carried by the existing `ruff-check` pre-commit hook (whole-tree, `pass_filenames: false`) and the CI lint job. Runtime enforcement is `filterwarnings = ["error::EncodingWarning"]` plus `PYTHONWARNDEFAULTENCODING=1` on the CI test job, which catches the fixture-derived receivers PLW1514 cannot type.
2. Pre-commit and CI enforcement are both preserved: pre-commit via the ruff hook, CI via both the lint job and the test job.
3. `tools/check_encoding.py` and its `check-encoding` hook are gone, and the directory is now empty.
4. Every reference is updated: the "Explicit encoding" convention is rewritten to name both mechanisms, both silent-disable dependencies (ruff `preview` and the env var), the blind spot, the residual, and the sentinel; the "### Tools" section is dropped; both documented `pytest` commands carry the env var; `docs.yaml` drops the `Tools:` heading.

### Coherence

1. The deletion drops no invariant the new mechanisms don't keep. The tree is compliant: of 239 text read/write calls in the suite, every one carries `encoding=` except the sentinel's deliberate omission, so the new error-gate passes cleanly rather than masking existing violations.
2. The sentinel does real work. With the env var unset it fails loudly ("DID NOT WARN"); with it set it passes. `pytest.warns` installs its own filter inside the block, so the `error::EncodingWarning` promotion does not interfere.
3. The residual is exactly the accepted one: a fixture-derived write on a never-executed line.
4. The generated index (`namespace.yaml`, `docs.yaml`, the new stub) is in sync, and no stale entry for the deleted script remains.

### One note, not a defect

The "Explicit encoding" convention still names `tools/` in the rule's scope though the directory is now empty. This is the deliberate whole-tree scope from the accepted design: a future `tools/` script is covered by PLW1514 statically. It is intentional, not a stale reference.

### Out of scope but noticed

The static half has a silent-disable vector with no guard: removing `preview = true` from the ruff config would drop PLW1514 with only a documentation note to catch it, unlike the runtime half, which the sentinel guards. This was a conscious, accepted decision at design time (documentation judged proportionate).

alimanfoo:
## Response to the reviews

Thanks to both reviewers. How each finding was handled.

**Sentinel guards the env var but not the filter escalation (non-blocking 1): accepted.** The sentinel now uses `pytest.raises(EncodingWarning)`, so it goes red if either the env var is unset or the `filterwarnings` escalation line is removed (commit `4fc4f98`). All three states verified. The doc was updated to credit both vectors (`6502d4c`).

**Two unguarded config lines (non-blocking 2 / "out of scope but noticed"): split.** The `filterwarnings` half is now guarded by the strengthened sentinel above. The `preview = true` half is left to documentation deliberately: a guard there would have to test the linter itself rather than the product, and removing preview would drop more than this one rule. Both silent-disable dependencies are named in the convention.

**"Two mechanisms" states a count it then lists (nit 1): accepted.** Reworded (`340d08d`).

**`tools/` named in the scope though the directory is empty (nit 2 / the one note): accepted.** The scope line now reads "the repository" rather than enumerating roots, keeping the whole-tree intent without the orphaned mention (`340d08d`).

**Goal framing centres `EncodingWarning` but the design makes `PLW1514` primary (readability 1): explained.** The delivered design pairs the static `PLW1514` rule (whole tree) with the runtime `EncodingWarning` gate (the fixture-derived receivers `PLW1514` cannot type). The Design comment above sets this out, and the description's key-consideration note now points to it.

**"Preserved behaviour" reads absolute but a residual exists (readability 2): accepted.** The narrowing was a conscious accept at scope. The description's preserved-behaviour wording now names the residual so it no longer contradicts the convention doc.

The completeness and coherence confirmations needed no action.

Issue #226, opened 2026-06-25
Complexity counters may accrue noqas instead of catching defects


alimanfoo:
Noticed during #225. The phase-5 complexity counters (PLR0911, PLR0912, PLR0913, PLR0915, PLR1702, enabled in #212) may accumulate suppressions on idiomatic dispatch code rather than catch real complexity.

They produced no retained fixes this session and one immediate suppression: `_check_call` in `tools/check_encoding.py` tripped PLR0911 with eight returns and now carries a `# noqa`, though it is a flat one-return-per-case dispatch that `C901` already accepts. Whether the counters earn their place is worth confirming over the next few sessions. Relates to #207 and #216.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

alimanfoo:
From the #225 retrospective: if the counters stay, the repo should also state how to judge a PLR09xx violation. CLAUDE.md covers C901 ("refactor rather than raise the threshold") but says nothing about PLR0911/0912/0913/0915. When the PLR0911 noqa landed this session, there was no stated rule to judge suppress-versus-refactor against. Extending that CLAUDE.md paragraph to the PLR09xx family would close the gap.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

alimanfoo:
Noticed during #232. The cognitive-complexity gate (complexipy) now exists, which gives a concrete lens for this question: cognitive complexity is the nesting-weighted metric that `PLR0912` (too many branches) and `PLR1702` (nested blocks) approximate, so their redundancy against it is now assessable. #232 also generalised the violation-judgement rule in `AGENTS.md` to all enabled complexity checks. The question this issue raises — whether the counters earn their place — is unchanged; the new tool just makes it answerable. The same question arguably applies to `C901` itself.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #204, opened 2026-06-25
Add ruff complexity check (C901)

alimanfoo:
Adds ruff's `C901` (mccabe) complexity check at `max-complexity = 8`, and refactors the functions it flags so the check passes clean across the whole tree. Addresses #202.

## Requirements Analysis

Session Type: maintenance.

### Improvement goals

1. ruff selects `C901` with `[tool.ruff.lint.mccabe] max-complexity = 8`, added together, as a stable rule (not in the preview block).
2. Every function ruff lints scores 8 or below, with no ignore list. ruff lints the whole tree (`src`, `tests`, and `tools`), not just `src`. Six functions exceeded 8 and were brought down by behaviour-preserving decomposition: `extract_module` (`src/uncoded/extract.py`), `extract_headings` (`src/uncoded/docs_map.py`), `_write_stubs` and `render_stub` (`src/uncoded/stubs.py`), `_sync` (`src/uncoded/cli.py`), and `check_file` (`tools/check_encoding.py`).

### Preserved behaviour

The refactors change structure only, never behaviour. uncoded's generated artefacts (namespace map, stubs, docs map, skill files) and CLI output stay identical, and the encoding checker's output is unchanged. The consumers who rely on that are the contributors and AI agents who run the pipeline, plus anything downstream that reads the generated index.

### Note

The enforcement surface was widened during the session. The initial analysis measured `ruff check src tests` and concluded the work touched `src` only. Enabling the check showed ruff lints the whole tree, including `tools/`, where `check_file` scored 14. The complexity ceiling holds wherever ruff runs, so `tools/` was brought under it too, with no ignore list.

Closes #202

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #232, opened 2026-06-26
Add cognitive complexity check (complexipy)

### LINKED ISSUE #216: Add cognitive complexity check (complexipy)
Possible follow-on to #207. Complements ruff's cyclomatic complexity (#202) with cognitive complexity,
which ruff does not implement. Separate tool, pre-commit only.

https://github.com/rohaquinlop/complexipy

#202 adds ruff's `C901` (cyclomatic complexity). Ruff does not implement
**cognitive** complexity — nesting-weighted, closer to how hard code is to
read — which is squarely on-mission for uncoded. `complexipy` (Rust, fast)
fills that gap and slots into the existing pre-commit + CI-via-pre-commit
setup the same way the other hooks do.

Add to `.pre-commit-config.yaml` alongside the ruff hooks:

```yaml
- repo: https://github.com/rohaquinlop/complexipy-pre-commit
  rev: v5.1.0    # pin to current; keep fresh with pre-commit autoupdate
  hooks:
    - id: complexipy
      exclude: ^tests/        # complexity in tests matters less
      # default threshold is 15; tighten when ready, e.g.:
      # args: [--max-complexity-allowed, "10"]
```

Because CI runs `pre-commit run --all-files`, this hook gates CI automatically
— no separate workflow step needed. (This is why I'm *not* proposing
complexipy's documented `--diff main --ratchet` CI recipe: ratchet is for
regression-only gating on codebases too large to fix up front, and it fits the
all-files pre-commit model poorly. The codebase is small and held to 100%
coverage, so an absolute threshold and fixing any current offenders is cleaner.
If we'd rather not fix existing offenders, fall back to ratchet in a dedicated
CI step instead of this hook.)

Suggested implementation steps:

- add complexipy-pre-commit hook to `.pre-commit-config.yaml`
- run `uvx complexipy .` to see current scores; pick a threshold
- refactor any functions over threshold (or `# complexipy: ignore` with justification)
- cross-reference the CLAUDE.md complexity note added in the ruff issue
- `uv run pre-commit run --all-files` green

alimanfoo:
Adds a cognitive-complexity check to the pre-commit suite and refactors the functions that exceed it. Addresses #216.

Closes #216

## Requirements Analysis

Session Type: maintenance. It adds a static check (cognitive complexity, via complexipy) and refactors the functions that exceed it. Behaviour stays the same. It follows #204, which added the C901 cyclomatic check.

### Improvement goals

Each goal is a checkable property of the code.

1. A cognitive-complexity check (complexipy) runs at pre-commit, and so in CI, failing when any covered function scores above 15.
2. The check covers tests as well as source, matching the scope of the existing C901 cyclomatic check.
3. Every covered function scores at or below 15. The four current offenders are refactored, with no `# complexipy: ignore` suppression left behind: `_find_config_file` (21), `sync_skills` (17), `_extract_params` (17), `resolve_ast_node_from_source` (16).
4. The conventions section in `AGENTS.md` names the cognitive-complexity check beside the existing cyclomatic ceiling, and states how to judge a violation: refactor or flatten, never suppress or raise the threshold. This matches the C901 rule and closes the gap #226 raised about complexity counters accruing suppressions.
5. The hook is pinned to a current release, consistent with the repo's other pinned pre-commit hooks.

### Preserved behaviour

Refactoring the offenders preserves their behaviour. uncoded's generated outputs (`namespace.yaml`, stubs, `docs.yaml`, skills) and CLI behaviour are unchanged. The suite stays green at 100% branch coverage.

### Decided shape

An absolute threshold gated through the existing pre-commit hook, not complexipy's `--diff main --ratchet` CI recipe. The issue rejects ratchet as a poor fit for a small, fully-covered codebase run all-files.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -32,6 +32,11 @@ repos:
         entry: uv run ty check src tests
         language: system
         pass_filenames: false
+      - id: complexipy
+        name: complexipy
+        entry: uv run complexipy src tests
+        language: system
+        pass_filenames: false
       # Runs before uncoded so the generated skill files are built from
       # already-formatted sources. The generated skill dirs are excluded
       # because uncoded sync owns those files.
--- a/.uncoded/namespace.yaml
+++ b/.uncoded/namespace.yaml
@@ -114,6 +114,7 @@ src/:
       _skill_path:
       _render_content:
       _remove_skill_file:
+      _sync_one_skill:
       sync_skills:
     stubs.py:
       VALUE_WIDTH_CAP:
--- a/.uncoded/namespace.yaml
+++ b/.uncoded/namespace.yaml
@@ -100,6 +100,7 @@ src/:
       UnsupportedNamePathError:
       resolve_ast_node:
       resolve_name_position:
+      _top_level_match:
       resolve_ast_node_from_source:
       _resolve_class_member:
     skill.py:
@@ -197,6 +198,7 @@ tests/:
       test_shadowed_class_member_found_in_last_definition:
       test_shadowed_class_member_not_found_in_last_definition:
       test_not_found_in_class:
+      test_class_member_found_when_function_has_same_head_name:
     TestUnsupportedNamePathError:
       SUPPORTED_SHAPES:
       _assert_raises:
--- a/.uncoded/namespace.yaml
+++ b/.uncoded/namespace.yaml
@@ -142,6 +142,8 @@ src/:
         constants:
         classes:
         functions:
+      _unparse_or_none:
+      _args_to_params:
       _extract_params:
       _render_value:
       _extract_assignment:
--- a/.uncoded/namespace.yaml
+++ b/.uncoded/namespace.yaml
@@ -34,6 +34,7 @@ src/:
         doc_roots:
       find_pyproject_toml:
       _has_uncoded_section:
+      _config_file_at:
       _find_config_file:
       read_config:
     docs_map.py:
--- a/.uncoded/stubs/src/uncoded/config.pyi
+++ b/.uncoded/stubs/src/uncoded/config.pyi
@@ -11,6 +11,9 @@ def find_pyproject_toml(start: Path) -> Path | None:
 def _has_uncoded_section(*, pyproject_path: Path) -> bool:
     ...
 
+def _config_file_at(*, directory: Path) -> Path | None:
+    ...
+
 def _find_config_file(*, start: Path) -> Path | None:
     ...
 
--- a/.uncoded/stubs/src/uncoded/resolver.pyi
+++ b/.uncoded/stubs/src/uncoded/resolver.pyi
@@ -13,6 +13,9 @@ def resolve_ast_node(name_path: NamePath, in_path: Path) -> ast.stmt:
 def resolve_name_position(name_path: NamePath, in_path: Path) -> tuple[int, int]:
     ...
 
+def _top_level_match(*, node: ast.stmt, head: str, tail: str | None) -> bool:
+    ...
+
 def resolve_ast_node_from_source(*, name_path: NamePath, source: str, in_path: Path) -> ast.stmt:
     ...
 
--- a/.uncoded/stubs/src/uncoded/skill.pyi
+++ b/.uncoded/stubs/src/uncoded/skill.pyi
@@ -21,6 +21,9 @@ def _render_content(*, skill: Skill) -> str:
 def _remove_skill_file(*, path: Path, project_root: Path, check: bool) -> bool:
     ...
 
+def _sync_one_skill(*, skill: Skill, source: bool, docs: bool, project_root: Path, check: bool) -> int:
+    ...
+
 def sync_skills(*, source: bool, docs: bool, project_root: Path, check: bool) -> int:
     ...
 
--- a/.uncoded/stubs/src/uncoded/stubs.pyi
+++ b/.uncoded/stubs/src/uncoded/stubs.pyi
@@ -11,6 +11,12 @@ from uncoded.sync import remove_file, sync_file
 
 VALUE_WIDTH_CAP = 80
 
+def _unparse_or_none(*, node: ast.expr | None) -> str | None:
+    ...
+
+def _args_to_params(*, arg_nodes: list[ast.arg]) -> list[StubParam]:
+    ...
+
 def _extract_params(args: ast.arguments) -> list[StubParam]:
     ...
 
--- a/.uncoded/stubs/tests/test_body.pyi
+++ b/.uncoded/stubs/tests/test_body.pyi
@@ -76,6 +76,9 @@ class TestResolveBodyClassMember:
     def test_not_found_in_class(self, tmp_path):
         ...
 
+    def test_class_member_found_when_function_has_same_head_name(self, tmp_path):
+        ...
+
 class TestUnsupportedNamePathError:
     SUPPORTED_SHAPES = ("'name'", "'Class/member'")
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -106,8 +106,9 @@ pull request and will fail a build where a hook was skipped.
 Do not pass `--unsafe-fixes` unless a specific violation needs it and you have
 reviewed the change.
 
-A C901 violation means a function is too complex to pass the check. Refactor or
-flatten it rather than raising the `max-complexity` threshold.
+A complexity-check violation means a function is too complex to pass. Refactor
+or flatten it. Never suppress a violation with an ignore comment and never raise
+the threshold. The enabled checks and their thresholds are in `pyproject.toml`.
 
 The ty pre-commit hook runs the type checker. It is pinned via the `dev`
 optional dependency, the same way ruff is.
@@ -158,9 +159,9 @@ Run tests locally as `PYTHONWARNDEFAULTENCODING=1 uv run pytest`. The sentinel
 in `tests/test_encoding_gate.py` fails loudly if either the env var is unset or
 the filterwarnings escalation is removed.
 
-**Complexity ceiling.** The value is in `[tool.ruff.lint.mccabe]` in
-`pyproject.toml`. See [Linting and formatting](#linting-and-formatting) for the
-response to a C901 violation.
+**Complexity ceiling.** The enabled complexity checks and their thresholds are
+in `pyproject.toml`. See [Linting and formatting](#linting-and-formatting) for
+the response to a violation.
 
 **Index committed.** Commit `.uncoded/` and keep it current with the pre-commit
 hook. See [Commands](#commands) and [Dev setup](#dev-setup).
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -107,8 +107,9 @@ Do not pass `--unsafe-fixes` unless a specific violation needs it and you have
 reviewed the change.
 
 A complexity-check violation means a function is too complex to pass. Refactor
-or flatten it. Never suppress a violation with an ignore comment and never raise
-the threshold. The enabled checks and their thresholds are in `pyproject.toml`.
+or flatten it. Never suppress a complexity violation with an ignore comment and
+never raise the threshold. The enabled checks and their thresholds are in
+`pyproject.toml`.
 
 The ty pre-commit hook runs the type checker. It is pinned via the `dev`
 optional dependency, the same way ruff is.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,6 +23,7 @@ classifiers = [
 
 [project.optional-dependencies]
 dev = [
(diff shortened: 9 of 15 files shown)


alimanfoo:
## Code Analysis

### The check-infrastructure layer

`.pre-commit-config.yaml` runs two hook styles. `markdownlint-cli2` and `pre-commit-hooks` are remote pinned repos (`rev: v...`). `ruff-check`, `ruff-format`, `ty`, and `uncoded` are `local` system hooks that shell to `uv run`, so their version is pinned once via the `dev` optional dependency rather than a `rev`. complexipy is not yet a dependency (absent from `pyproject.toml` and `uv.lock`). So a hook "pinned consistent with the other hooks" has two existing patterns to choose between.

`pyproject.toml` `[tool.ruff.lint]` (line 67) holds the existing complexity gate: `C901` in the `select` list (line 77), `[tool.ruff.lint.mccabe] max-complexity = 8` (lines 110-111), and `preview = true` (line 68). The cyclomatic ceiling already lives here.

`[tool.ruff.lint.per-file-ignores]` (line 116) sets `"tests/**" = ["D", "ANN", "S101"]`. It does not exempt `C901`, so cyclomatic complexity already applies to tests. That is the precedent for covering tests with the new check.

`.github/workflows/ci.yml` runs `uv run pre-commit run --all-files` (line 19). Any pre-commit hook gates CI with no extra workflow step.

### The conventions documentation

`AGENTS.md` has a "Conventions" section with a "Complexity ceiling" entry (line 161) that names `[tool.ruff.lint.mccabe]` and points to "Linting and formatting". That section (line 109) states the response to a `C901` violation: refactor or flatten rather than raise the threshold. The repo carries no coverage pragmas and, per #226, treats accumulated suppressions as a smell. This is the rule to extend to cognitive complexity.

### Why the check adds something

All four offenders currently pass `C901` (cyclomatic at or below 8, since CI is green) yet exceed cognitive 15. Cognitive complexity is nesting-weighted, so it flags functions that read as hard but branch few times. That gap is what this change targets.

### The four offenders

Each works and is covered. None signals a behaviour defect. The smell in each is that it folds two or more concerns into one body.

1. `_find_config_file` (`config.py`, score 21). A `while True` walk up the directory tree wraps a four-way per-directory decision (pyproject-with-section, `.uncoded.toml`, both, neither). The walk and the per-directory rule are two concerns nested into one function. The docstring already states the per-directory rules as a distinct list.
2. `sync_skills` (`skill.py`, score 17). A loop over `SKILLS` nests a build-or-remove branch, each arm looping over `SKILL_ROOTS`, plus a third nested loop over legacy names by roots. The nesting drives the score; the per-skill work and the legacy cleanup are separable concerns.
3. `_extract_params` (`stubs.py`, score 17). A flat sequence of blocks, one per argument kind. The idiom `annotation = ast.unparse(arg.annotation) if arg.annotation else None` repeats five times: a small duplicated fact (how to read an annotation off an arg node) restated at each block.
4. `resolve_ast_node_from_source` (`resolver.py`, score 16). A loop over top-level AST children holds a four-branch `if/elif` chain matching the head name against `ClassDef`, `FunctionDef`, `Assign`, `AnnAssign`, and `TypeAlias`, tracking the match, then dispatches to class-member resolution. The match decision and the walk are nested into one function.

alimanfoo:
## Scope

1. A cognitive-complexity check runs in the pre-commit suite, and so in CI, failing when any covered function — source or tests — scores above 15.
2. Every covered function scores at or below 15. The four offenders are refactored, with no suppression left behind.
3. The conventions section in `AGENTS.md` states, once and generally, how to judge a complexity-check violation: refactor or flatten, never suppress or raise the threshold. It covers the cognitive check together with the already-enabled cyclomatic (`C901`) and `PLR09xx` counters, and points to where the enabled set and thresholds live rather than restating any code list or number.
4. The check's version is pinned, consistent with the repo's pinned hooks.

The judgement-rule coverage in item 3 documents the counters enabled today. It does not decide #226's open question of whether those counters should stay.

alimanfoo:
## Design

### The gate

- Add `complexipy` (5.6.1) to the `dev` optional dependency in `pyproject.toml`, pinned there like `ruff` and `ty`.
- Add a `local` pre-commit hook running `uv run complexipy src tests`, beside `ruff` and `ty`. CI already runs `pre-commit run --all-files`, so no `ci.yml` change.
- Put the threshold in `pyproject.toml` `[tool.complexipy]`, `max-complexity-allowed = 15`, beside `[tool.ruff.lint.mccabe]`. One home for the number; complexipy reads it there.
- Source and tests are both covered, via the `src tests` path args, mirroring `ty`.

### The refactors

All four are behaviour-preserving.

- `_find_config_file`: replace the manual `while True` walk and `parent == current` sentinel with iteration over the start directory and its parents, and extract a private single-directory probe helper that returns `Path | None` and raises `ConfigError` on the ambiguous same-directory case. The helper is named for the probe, not "find", so its "nothing in this directory" `None` is not confused with the walk's "reached the root" `None`. Lower nesting cuts the score directly.
- `_extract_params`: extract `_unparse_or_none(node)` and route all seven idiom sites through it — the five here plus the return-annotation sites in `_extract_function` and `_property_attribute` — so the fact has one home. Extract `_args_to_params` to collapse the three identical argument-list loops; the `/` and `*` markers stay in the caller.
- `sync_skills`: extract a per-skill helper covering one skill's build-or-remove plus its legacy cleanup. The `SKILLS` loop body becomes one call.
- `resolve_ast_node_from_source`: extract a `_top_level_match` helper, carrying the tail-is-`None` condition in its signature (required for the function, assign, and type-alias arms, not the class arm) so the match is not silently broadened. The loop keeps the match; class-member dispatch follows.

### The documentation

Generalise both the "Complexity ceiling" entry (which currently says the value is in `[tool.ruff.lint.mccabe]`, wrong once there are two ceilings) and the "Linting and formatting" paragraph it points to (which names only `C901`). State one rule over all enabled complexity checks: refactor or flatten, never suppress or raise. Point to `pyproject.toml` for the live enabled set and thresholds, restating no list or number.

### Alternatives considered

- Remote pinned `complexipy-pre-commit` hook, like `markdownlint`. Trade-off: a second tool-install path (the hook's own environment) that can drift from the version a developer runs through the dev dependency. The local hook single-sources the version.
- Threshold as a CLI argument in the hook entry. Trade-off: the number then lives in `.pre-commit-config.yaml`, a second home, rather than beside the other complexity config in `pyproject.toml`.

alimanfoo:
## Plan

The refactors land before the gate, so the gate passes the moment it is added. The documentation comes last.

1. Refactor `_find_config_file` (`config.py`), behaviour-preserving. Replace the manual upward walk with iteration over the start directory and its parents, and extract a private single-directory probe helper returning `Path | None` (raising `ConfigError` on the ambiguous same-directory case). The helper is named for the probe, so its "nothing in this directory" `None` stays distinct from the walk's "reached the root" `None`.
2. Refactor `sync_skills` (`skill.py`), behaviour-preserving. Extract a per-skill helper covering one skill's build-or-remove plus its legacy cleanup; the loop body becomes one call.
3. Single-source the unparse-optional-annotation idiom in `stubs.py`, behaviour-preserving. Extract `_unparse_or_none(node)` and route every occurrence of `ast.unparse(X) if X else None` through it. Extract a helper for the identical argument-list loops; the `/` and `*` markers stay in the caller.
4. Refactor `resolve_ast_node_from_source` (`resolver.py`), behaviour-preserving. Extract a `_top_level_match` helper that carries the tail-is-`None` condition in its signature (the function, assign, and type-alias arms require it; the class arm does not), so the match is not broadened.
5. Add the cognitive-complexity gate. Add complexipy to the dev optional dependency; set `[tool.complexipy] max-complexity-allowed = 15` beside `[tool.ruff.lint.mccabe]`; add a local pre-commit hook running complexipy over `src` and `tests`, beside the ruff and ty hooks. CI already runs the whole pre-commit suite.
6. Generalise the complexity-convention documentation in `AGENTS.md`. State one rule over all enabled complexity checks: refactor or flatten, never suppress or raise the threshold. Point to `pyproject.toml` for the live enabled set and thresholds rather than restating any code list or number.

alimanfoo:
## Review (fresh-eyes)

**Recommendation:** looks good, a couple of small non-blocking notes below.

This is a clean PR. The four refactors are strict extract-helper moves with no behaviour change: I read each one, and two parallel checks confirmed the config walk, the resolver matcher, the stubs parameter builder, and the skill loop all produce identical results to the originals. The new complexity gate is wired the same way as the existing `ty` hook and reads its threshold from `pyproject.toml`. My cold read of the diff matched the stated intent with no divergence of substance.

### Non-blocking

1. AGENTS.md, the new sentence "Never suppress a violation with an ignore comment", reads as an absolute rule but is meant only for complexity violations. The repo already carries four justified `noqa: S404/S603/S607` security suppressions (`src/uncoded/refs.py` lines 7, 67, 68 and `tests/test_refs.py` line 3). An agent reading this instruction literally could flag or strip those legitimate suppressions. The sentence sits in the complexity paragraph, so a careful human infers the scope, but the codebase is read by agents that act on every word. Narrowing it to "Never suppress a complexity violation with an ignore comment" removes the contradiction without changing the intent.

### Nits

1. `pyproject.toml`, `[tool.complexipy] max-complexity-allowed = 15`: this equals complexipy's own default, so the line documents the default rather than tightening anything. That is fine and arguably clearer for a reader, but worth knowing that the meaningful knob is only engaged if the value drops below 15.
2. `src/uncoded/config.py`, `read_config` docstring: it dropped the pointer to where the precedence rules live. Those rules moved to `_config_file_at`, so a reader of `read_config` is no longer directed to them. A short "see `_config_file_at`" would restore the trail.

### Readability

1. The "never suppress with an ignore comment" clause is technically accurate in its paragraph, but the diff alone does not reveal that the rule is scoped to complexity. A cold reader cannot tell whether it governs all checks or just the two complexity gates, which is the root of the non-blocking note above.

alimanfoo:
## Review (requirements and scope)

**Recommendation: ready to merge.** The change delivers the agreed scope in full, the four refactors are behaviour-preserving, the new gate is wired and pinned correctly, and I found nothing that blocks. No findings need action before merge.

What I checked across the whole diff:

1. The gate runs and bites. `complexipy` runs as a local pre-commit hook over `src` and `tests`, so it also runs in CI through `pre-commit run --all-files`. The threshold lives once in `pyproject.toml` under `[tool.complexipy]`, beside the cyclomatic ceiling. A function above 15 fails the hook.

2. Every offender is genuinely reduced, not suppressed. `_find_config_file`, `sync_skills`, `_extract_params`, and `resolve_ast_node_from_source` are each refactored by extracting a focused helper. No `# complexipy: ignore` or `# noqa` was added anywhere in the diff.

3. The refactors preserve behaviour. I confirmed each: the config walk visits the same directories in the same order as the old loop; the resolver predicate keeps the class-versus-function tail rule exactly, and a new test pins it against a future regression; the annotation idiom now has one home that expands to the same expression at every site; the per-skill helper keeps the same change-count semantics.

4. The convention doc is generalised, not enumerated. Both the "Linting and formatting" paragraph and the "Complexity ceiling" entry now state one rule for any complexity check, refactor or flatten and never suppress or raise, and point to `pyproject.toml` for the live set and thresholds. This is the intended shape: no code list or number is restated, so the doc cannot drift from the config.

5. The generated index is consistent. All five new helpers appear in the namespace map and stubs in source order; no generated file is stale.

One note for a reader who expects the doc to name each check: naming them was deliberately avoided in favour of a single rule that points to `pyproject.toml`, so a future check is covered without editing the prose.

Out of scope but noticed:

- `find_pyproject_toml` in `config.py` still uses the older `while True` walk that `_find_config_file` moved away from. The two now express the same upward walk two ways in one file. Pre-existing and independent of this change; worth a small follow-up to align them.

alimanfoo:
## Response to the reviews

**Fresh-eyes review**

- Non-blocking 1 / Readability 1 (the no-suppress sentence read as an absolute ban): accepted and fixed in d0a6e8c. The sentence now reads "Never suppress a complexity violation with an ignore comment", so it no longer reads as banning the repo's justified `noqa: S...` security suppressions.
- Nit 1 (the threshold equals complexipy's default): left as is, deliberately. Pinning the value explicitly guards against a future change to the tool's default and gives the convention doc one place to point at.
- Nit 2 (`read_config` no longer points to the precedence rules): left as is. `read_config` states its own public contract (what it raises and returns), and the per-directory precedence rules are an implementation detail of the private `_config_file_at`. A public docstring pointing into a private helper was removed on purpose earlier in the branch.

**Requirements-and-scope review**

- Ready to merge, no action needed.
- Out of scope (`find_pyproject_toml` still uses the older `while True` walk): held for post-merge triage. It is pre-existing and independent of this change.

